PR #1, opened 2022-06-03
Fix the runApp() usage during `jaspr create my_web_app`

Clean-Cole:
Creating a new `jaspr` web app with the pub executable creates an application with initial errors.

The following line does not properly build or compile.

```dart
runApp(() => App(), id: 'output');
```

But works properly when changed to 

```dart
runApp(App());
```

P.S. Thank you for making this package.


schultek:
Thanks

schultek:
Published in v0.1.1

PR #9, opened 2022-06-08
Add windows support

slightfoot:


packages/jaspr/bin/jaspr.dart
```diff
@@ -274,7 +274,7 @@ Future<Process> _runWebdev(List<String> args) async {
   Process process;
 
   if (hasGlobalWebdev) {
-    process = await Process.start('webdev', args);
+    process = await Process.start(Platform.isWindows ? 'webdev.bat' : 'webdev', args);
```
schultek: Maybe `dart pub global run webdev` would be better here.

slightfoot: I like that. Good idea. Will give it a go.

PR #22, opened 2022-07-15
v0.2: jaspr ui: components + html + styling

### LINKED ISSUE #16: Add utility methods / components for common html elements
Discussion related to #13 

## Status quo

The current way of defining html elements is by using `DomComponent` and providing a `tag`. This leaves the most freedom to developers, but can get verbose quickly when composing uis:

This: 
```dart
return DomComponent(
  tag: 'div',
  children: [
    DomComponent(
      tag: 'h1', 
      child: Text('This is a title'),
    ),
    DomComponent(
      tag: 'p',
      children: [
         Text('Hello '),
         DomComponent(
           tag: 'b',
           child: Text('World!'),
        ),
      ],
    ),
  ],
);
```

is unarguably a lot more verbose than standard html:
```html
<div>
  <h1>This is a title</h1>
  <p>Hello <b>World!</b></p>
</div>
```

## Proposal / Discussion

While `DomComponent` should stay as the low-level base for all html elements, we need a more simple way to use common html elements like `div` and others. This should be added on top of the core framework and be optional to use by the developer. 

### What should such an element look like?

There are two ways we could define such elements:

1. As their own `StatelessComponent`
2. As methods returning a `DomComponent`

Separate components (1) would define a class for each element, take any arguments in the constructor and return a `DomComponent` in the `build()` method. Since classes use *PascalCase* for their naming, we either 
- (a) capitalize the tag name (`Div`, `P`, `H1`), or 
- (b) extend the tags to their more descriptive names (`Division`, `Paragraph`, `Heading1`).  

Both are not really ideal, since (a) would have a bunch of single-letter class names and (b) would be again quite verbose for a lot of elements. Additionally, having separate components introduces a lot of extra depth into the component tree when building uis (2 components for each element, the elements component + the dom component), which will have some performance implications for larger apps.

Element methods (2) would provide a way to define shorthands for common elements without the overhead of additional components. A method can also have custom properties specific for the element and would just return a single `DomComponent` with the static tag and provided attributes. Methods also have lower-case names and could directly mirror the original html tags (`div()`, `p()`, `h1()`), keeping the code concise and appearing similar to original html.

### Which elements are included?

Of course we want as many elements included as possible, however this might be a challenge as there are a lot of different elements in the html spec. 

When writing this manually we would have to start with a few most common elements and expand the list over time. However this would also mean a lot of maintenance when the component api or properties of `DomComponent` change (which might happen since jaspr is very early stage).

A better and more easy way would be to generate the methods and their bodies for each html element automatically. A good example would be [domino_html](https://github.com/agilord/domino/tree/main/domino_html) which does exactly this for the domino package. It uses a script that crawls [developer.mozilla.org/en-US/docs/Web/HTML/Element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element) and [developer.mozilla.org/en-US/docs/Web/HTML/Attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) to get all html elements and their attributes. Modifying this script to output jaspr-compatible code should be feasible. There are however some additional considerations to make:

- Some of the element have many uncommon attributes. These get very verbose and make the parameter list unclear and convoluted. A possibility would be to not include *global attributes* as defined by the attributes page.
- Some attributes have custom types other than `String` (mainly `boolean`, `int` or enums). It would be cool to provide them as typed attributes (especially the enums), but I don't know right now how to get this information, since it is not specified in the attributes page.

### LINKED ISSUE #18: Styling in jaspr

schultek:
This is to keep track of the changes made to the wip/v0.2 branch.

resolves #16, resolves #18 

TODOs:

- [x] add html utility components
- [x] update rendering implementation
- [x] run as shelf middleware
- [x] run multiple apps on client
- [x] finish islands architecture implementation
- [x] implement document component
- [x] add benchmarks
- [x] implement typed styles
- [ ] experiment with inherited styles
- [ ] implement common ui components
- [ ] add jaspr config in pubspec (execution mode)
- [ ] bump test coverage
- [ ] new router implementation

> Note to myself: These are getting out of hand, maybe split into two updates.😅


schultek:
@mjablecnik You can add a PR into this branch and I will merge from time to time to keep it updated. This PR is just to keep track of all combined changes.

juancastillo0:
Hi, thanks for your work

I have some Bootstrap components and types implemented in [this repo](https://github.com/juancastillo0/bootstrap_dart). They may be useful for implementing some of the work being done in the bootstrap folder. In particular, there are some [`@JS()` annotated classes](https://github.com/juancastillo0/bootstrap_dart/blob/main/lib/bootstrap/js_bindings.dart) like `Modal` with the corresponding separate interface for usage in non-web (server-side) environments. This is using a fork of [`package:deact`](https://pub.dev/packages/deact). However, I believe it could be ported to jaspr or at least use/explore a similar API. You can view the implemented components in [this web page](https://juancastillo0.github.io/bootstrap_dart) and [the code](https://github.com/juancastillo0/bootstrap_dart/tree/main/web) of the deployed page.

If there is anything I can help you with please let me know.

schultek:
Replaced by #44

Issue #26, opened 2022-07-29
`build` method middleware or wrappers to support MobX, hooks, Pods and other framework-wide functionalities


juancastillo0:
Hi, thanks for your work!

This may be a weird feature request which will make the package a bit different from Flutter and there may be performance considerations. However, the usability improvements would be great, in my opinion.

The idea is to provide a middleware for rendering Components, a function that wraps the executions of all build methods, to do various things.

For example, for registering and tracking dependencies that rebuild the Widget when they change. I believe the javascript framework https://www.solidjs.com/ is based on this concept, similar to MobX's tracking of dependencies. Through testing the package I have implemented some MobX and hooks bindings for Components. In [this file](https://github.com/juancastillo0/cidart/blob/0bb72153546754e99f8a024c46d337e81de7b867/web_client/lib/mobx_hooks.dart) you can see the main hooks logic and also the tracking of dependencies within build methods of `StatelessObserverComponent` similar to the one in [`package:flutter_mobx`](https://pub.dev/documentation/flutter_mobx/latest/flutter_mobx/StatelessObserverWidget-class.html), however this one also supports hooks. In Flutter this is a problem it's difficult to use `package:flutter_mobx` and `package:flutter_hooks` (two packages implementing Elements) without nesting that the composing of middleware or multiple wrappers can solve. There are no automatic test, but I have tested them manually and you can try the deployed html in [gh-pages](https://juancastillo0.github.io/cidart/). More hooks, using the same infrastructure, are implemented in [this hooks file](https://github.com/juancastillo0/cidart/blob/0bb72153546754e99f8a024c46d337e81de7b867/web_client/lib/hooks.dart). 

[...]

PR #27, opened 2022-08-01
Observer component and MobX Hooks integration example app

juancastillo0:
Closes https://github.com/schultek/jaspr/issues/26

Hi,

As discussed in the issue, this PR implements an `ObserverComponent` within the framework. The code is similar to some of the sections for `InheritedComponent`s, there is a `_updateObservers` that assigns the `_observerElements` from the parent and registers the element if it itself is an `ObserverElement`. The `ObserverElement` methods `willRebuildElement`, `didRebuildElement` and `didUnmountElement` are already executed in the build process. At the moment, they are required to be implemented (do not have a default implementation). However, it could be best to leave an empty function implemented?

I created a `mobx_hooks_experimental` app within the experimental folder. It contains the same logic as before for hooks and mobx tracking, but using the new `ObserverComponent` infrastructure. The `MobXHooksObserverComponent` is used only once at the top of the component tree in `main.dart` and all child elements are being tracked (they can use hooks and they subscribe automatically to observables).

I think I would like to make some tests for the framework and for the mobx_hooks part too. What do you think we should do about the MobX and Hooks integration? At the moment it is some kind of example app, however, it could be a complete package.

Thanks!


juancastillo0:
Hi,

The last commits add tests, along with small changes, to the framework, and a test with some refactoring to the hooks part.

The last couple of changes include overriding the update method from `ObserverElement` to make sure the children are rebuilt. And now `_observerElements` is cleared in unmount to make sure the execution of `didUnmountElement` has the registered observers.
The tests perform updates to the Component tree and listen to the execution of the methods in `ObserverElement`. There is a stage in which there are two `ObserverComponents` listening together. All the methods, including `didUnmountElement` are executed in the test. The rebuilding of the tree is performed localized and globally.

In the hooks part, I did some refactoring so that the jaspr part is very lean, the hooks logic can be used in a Dart only context (there is a test with multiple hooks). The jaspr component uses all methods in its `ObserverElement`, you can view it in the `experiments/mobx_hooks/lib/mobx_hooks/jasper_observer.dart` file.

I believe most of the things are ready, specially on the framework part.

Please let me know what do you think and if there is something more I can do.

Thanks!

packages/jaspr/lib/src/framework/framework.dart
```diff
@@ -772,6 +796,11 @@ abstract class Element implements BuildContext {
           dependency.didRebuildDependent(this);
         }
       }
+      if (_observerElements != null) {
```
schultek:
```suggestion
      if (_observerElements != null && _observerElements!.isNotEmpty) {
```

schultek:
I changed the target branch to `develop`, please check that it works with the new rendering system

packages/jaspr/lib/src/framework/framework.dart
juancastillo0: Done in https://github.com/schultek/jaspr/pull/27/commits/263281b07efbb25ac46edb8eab09c8f3c24a90f8

juancastillo0:
Hi thanks for reviewing

I just committed the proposed changes to check `_observerElements!.isNotEmpty` before iterating and, with the new rendering system, I had to change the expected events in the tests for `ObserverComponent`. Before, willRebuildElement and didRebuildElement was not being executed with `TextElement` as argument, only the didUnmountElement. In the last phase of the test I change the text in the `Text` component and the willRebuild and didRebuild are not called with the `TextElement` as argument. They are called only once when the element is created.

Also the `TestComponentController.rebuild()` was not rebuilding the root component. I had to change it to (I did not commit this change):

```dart
extension PumpTestComponent on ComponentTester {
  Future<TestComponentController<T>> pumpTestComponent<T>(TestComponent<T> component) async {
    await pumpComponent(component);
    Element? testElem;
    binding.rootElement!.visitChildren((element) {
      element.visitChildren((element) {
        testElem ??= element;
      });
    });
    return TestComponentController(testElem!);
  }
}
```

And the `TestDomBuilder` was missing a `@override void skipContent(DomNode node) {}`.

schultek:
Ok thanks, I will look at the bugs you mentioned.

About the Text component, it technically does not rebuild or build at all, since it has no build method and no children. But I can make it that the observers are still called.

juancastillo0:
Yes, I believe it would make sense that it does not rebuild, like a cost Component or one whose reference is the same. However, having the build event as well as the unMount I think does makes sense and the behavior would be the same as a const Component. This is the current situation. It would be weird not having the build, since there is an unmount event afterwards, but maybe for that one should listen to the mount event and not the build one.

For hooks or MobX it is not that important since the most important elements to track are the ones that have build methods where one uses hooks or observables, but maybe other applications of ObservableComponent could use the other elements lifescycle.

Thanks again for reviewing!

schultek:
When you merge with develop the bugs with the testing library should be fixed.

I thought about the issue with Text and decided I don't want to artificially call the build hooks when the text component infact never rebuilds. As you said for most use-cases this shouldn't matter, since the main focus is to track the build methods of stateless and stateful components. If this is needed in the future this can easily be added, but for now I would just leave it as it is.

juancastillo0:
Great, thanks!

The test are working now. No issues in the merge.

PR #31, opened 2022-08-06
Fix small bugs

mjablecnik:


schultek:
Having the `List` instead of `Iterable` was a deliberate choice, I felt it was more stable since iterables would only be evaluated lazily and lists are not.
I think most of the time this will be just used with a `List` literal on `DomComponent`, which would not make a difference.
Do you have a use-case where `Iterable` would be preferred?

mjablecnik:
@schultek I only want to fix html functions in jaspr_html:
image

schultek:
I fixed this directly in the branch.

PR #56, opened 2023-02-01
Add flutter embedding demo

schultek:


schultek:
Currently the cli needs to create a dummy `index.html` file until https://github.com/flutter/flutter/issues/127487 is resolved.

PR #61, opened 2023-04-02
Fix GridPlacement styles being written to styles map

HankG:
GridPlacement style KV pairs weren't added to the generated GridItem styles because GridItem's styles property always returned an empty map. This returns the placement styles if they exist instead.


schultek:
Thanks

PR #70, opened 2023-04-05
Refactor List component

mjablecnik:
I've moved List into text components because it is in raw html and it looks more like text.
ListView with ListItem which looks like in Flutter I would like to create in near future.


schultek:
I don't think it should be tied to text semantically. You can put inside a list whatever you want and style it. So its more than just some bullet points.

mjablecnik:
Whole html is historically text document with paragraphs, lists, tables, etc. elements. After apply styles it can be transformed into non text components. But now it is only pure text list and nothing more.
In CSS frameworks like Bootstrap it can be called like ListView with card items but now it is only TextList where can you apply some nice styles.

schultek:
The name would suggest it is **only** for text, which it is not. Sure you can also build a list of cards using only divs, but that not the point.

schultek:
The name would suggest it is **only** for text, which it is not. Sure you can also build a list of cards using only divs, but that not the point.

mjablecnik:
So how do you would like to name it when `List` is named of core Dart list class? `ListView` I want to use for list of cards similar like in Flutter and `RawList` or `ListRaw` I don't know if it is a really good solution so `TextList` is according to me the best name.

easazade:
If it's a list that has Components for children, why not name it `ComponentList`?

schultek:
I don't get why it shouldn't be called ListView? Any list component should be independent of what children it has, whether text or cards or any other component.
In Flutter also all the ListView does is provide a scrollable list of children. These may be ListTiles, but also any other widget you want.

mjablecnik:
But in html there are used `<ul>`, `<ol>` and `<li>` tags.
In html when I use `<ul>` or `<ol>` list so inside these lists have to be `<li>` tags. There cannot be some other tags or not?
So there cannot be any Component.
If ListView will use `<div>` tags with some styles then there can be any component in children.

When I use `<ul>`, `<ol>` tags so result is this:
image

Ordered or unordered text list and therefore it should be named as `TextList` I think.

If I will create list with <div> elements then it can be named `ListTiles` or something similar but list in my PR doesn't use `<div>` elements..

schultek:
Since most html tags can be styled freely, choosing a html tag over another is mostly about the *semantics* of a layout. So its important for accessibility, screen readers, web crawlers etc..

A ul>li semantically means just a list of things. What things is irrelevant. It would be totally ok to use the list component to build a list of cards, images or whatever. So naming it TextList for me just feels wrong. 

I also wouldn't want to build another ListView component that uses divs, because of the semantics. It would still use ul. And then we would have two component that basically do the same thing.

I just dont see the value of this change. It creates more confusion than it solves.

easazade:
> Since most html tags can be styled freely, choosing a html tag over another is mostly about the _semantics_ of a layout. So its important for accessibility, screen readers, web crawlers etc..
> 
> A ul>li semantically means just a list of things. What things is irrelevant. It would be totally ok to use the list component to build a list of cards, images or whatever. So naming it TextList for me just feels wrong.
> 
> I also wouldn't want to build another ListView component that uses divs, because of the semantics. It would still use ul. And then we would have two component that basically do the same thing.
> 
> I just dont see the value of this change. It creates more confusion than it solves.

I agree. I know jaspr is primarily being developed for Flutter developers. But now or in the future, there will be other developers that will use jaspr. because of that, I think keeping things, styles, terms, code or names close to the web side instead of the Flutter side will be a benefit in long term.

mjablecnik:
Ok, so I renamed it and moved from text into list file.
Is it good right now?

MarkOSullivan94:
@mjablecnik would you be kind enough in helping me understand these classes by adding some documentation? 

Even a link to what this is (or something similar) in existing web components would be suffice and helpful, thanks! 😄

schultek:
I think we should circle back a bit and discuss how we want to design these components and work from there. Since they lay the groundwork for future stuff, like more opinionated component libraries, I'm very keen on having a good design here.

I will create an initial draft of a design doc over the next days.

Hope you agree @mjablecnik

PR #73, opened 2023-04-09
refactor: improved new project error when not using snake_case

### LINKED ISSUE #62: Create new project error
During creating of new Jaspr project should be showed some better error:

image

For example:
```
"JasprProject" is not a valid project name.
Project name should be in snake_case format.
```

Or automatically convert my project name into correct name of pubspec.yaml

MarkOSullivan94:
Resolves #62


easazade (commented):
👍

PR #124, opened 2023-09-27
Correcting the misinformation in the CLI mentioned in issue #123

### LINKED ISSUE #123: Wrong information in the CLI after using create command with client template
When creating a project using the CLI and the `jaspr create` command it outputs the information to next use:
`cd\DIRECTORY`
`jaspr serve`
It does not mention the necessary "--no-ssr" flag.
That is confusing and should be standardized.

LaconicDev:
Added a check to add the 'no-ssr' flag to the output information after creation, if necessary (=>client template).
Small function that is easily adaptable for other flags if necessary. This would unify the content of the created readme file and the information of the CLI.

Might be a smart idea to unify this logic with the template selection though, if the names might change. It basically double checks the name right now.
But I didnt want to interfere that far in the framework as i haven't worked with it yet.

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [x] 🔨 Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✏️ Documentation
- [ ] 🗑️ Chore

Fixes #123

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,3 +1,7 @@
+## unreleased 0.9.1
+
+- Corrected information returned by the CLI after using `jaspr create -t client` from `jaspr serve` to `jaspr serve -no-ssr`
+
 ## 0.9.0
 
 - Added *Static Site Generation* support.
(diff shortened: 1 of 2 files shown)


LaconicDev:
Noticed a mistake in the changelog

PR #126, opened 2023-10-04
Improvement: cli update

### LINKED ISSUE #121: feat: Add `jaspr doctor` command.
**Description**

Similar to `flutter doctor` to print out a bunch of info.

**Additional Context**

This is mainly for me, so I can ask people to put it in issues :)

schultek:
## Description

Improved the stability and logging of the cli and added the commands:
  - `clean` command to clean your project directory
  - `update` command to automatically update the cli to the latest version
  - `doctor` command to print information about the environment and project
  
I also added a lightweight anonymous usage tracking to the cli. I'm using mixpanel.com for this as they have a free tier and well designed dashboards. I plan to make the usage dashboard public, but will wait until this is released for a few days to get enough events to have real anonymity.

Resolves #121 

@dkbast DM me if you want a preview of the dashboard.


fusion44:
I'll update my local copy this weekend and give it a try. Looks like a great addition.

The only gripe I have as I scroll through the code is that there doesn't seem to be a way to opt out of the analytics. Especially as the project name is always logged.

schultek:
Cool thanks for testing.

You're right I forgot to add a way to opt out. Will do so.

The project id is hashed before its sent. I thought it might be a good metric to trac distinct projects in addition to distinct users.

schultek:
@fusion44 updated to allow disabling analytics

PR #141, opened 2023-11-19
feat: add a Nix flake file

fusion44:
This PR adds a Nix Flake file. This enables users of [NixOS or the Nix package manager](https://nixos.org/explore.html) to quickly setup a development environment without having to install the Flutter / Dart on the base system installation.

## Type of Change
- [ ] ❌ Breaking change
- [x] ✨ New feature or improvement
- [ ] 🛠️ Bug fix
- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore


fusion44:
Adding the file doesn't tie Jaspr to Nix in any way. This just a simple (optinal) way to create a shell environment without having to install Flutter or Node permanently on the base system when using Nix. Thats not easy to replicate with a shell script. You can also think of it like a Python environment but for system packages instead of Python packages. [Here's](https://nix.dev/tutorials/first-steps/ad-hoc-shell-environments#ad-hoc-envs) some additional info. 

This would save me (and other potential Nix users) some hassle when developing Jaspr itself since the file makes most sense in a projects root directory.

Anyway, I found a few problems I have to solve. Please don't mege yet.

schultek:
Hi, closing this as my original opinion still stands. If this would be part of the repo I would need to maintain it, which I don't want since I'm not using nix (which is funny because I'm german and in german 'nix' means 'nothing' ;D).

There are other ways to add this to the project though to help contributors to get started. My proposal would be this:

1. You put the config files in a public gist.
2. You open a discussion here that links the files explains how to use them.
3. After that I'm happy to refer to that discussion in the contributors guide.

PR #147, opened 2024-01-06
added StyleSheet object to link external stylesheets

cybertheory:
title

## Description

Added StyleSheet object to link external stylesheets.

- Added StyleSheet object to jaspr/lib/src/components/style.dart, so you can add custom CSS StyleSheets
- Modified _BaseDocument to inlcude option StyleSheet list attribute

## Type of Change

- [ ] ❌ Breaking change
- [x] ✨ New feature or improvement
- [ ] 🛠️ Bug fix
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

## Ready Checklist

- [x ] In case this PR changes one of the core packages, I've modified the respective **CHANGELOG.md** file using

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,8 +89,3 @@ pubspec_overrides.yaml
 # Conventional directory for build outputs
 build/
 
-# Jaspr folders and files
-components_test/
-example_app/
-NOTES.md
-
--- a/apps/jaspr_pad/lib/adapters/html.imports.dart
+++ b/apps/jaspr_pad/lib/adapters/html.imports.dart
@@ -1,8 +1,7 @@
 // GENERATED FILE, DO NOT MODIFY
 // Generated with jaspr_builder
 
-export '../generated/imports/_web.dart'
-    if (dart.library.io) '../generated/imports/_stubs.dart'
+export '../generated/imports/_web.dart' if (dart.library.io) '../generated/imports/_stubs.dart'
     show
         window,
         document,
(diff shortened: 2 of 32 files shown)


schultek:
Please run `melos format` before submitting a PR.

schultek:
This is not necessary. You can already do this:

```dart
Document(
  head: [
    link(rel: "stylesheet", href: "styles.css"),
  ],
  body: App()
)
```

cybertheory:
Also, one thing that confuses me is the separation of jaspr_core and components, even though they are both integrated in the same library. Is the plan to eventually migrate to strictly using BaseComponent's descendents?

schultek:
> should I make it a component instead, I think it's necessary, who wants to create a new link every time lol
And think they are common enough that they should get their parameter in Document without having to use head

Then I would make it actually just a parameter on document, like directly a list of strings.

> Also, one thing that confuses me is the separation of jaspr_core and components, even though they are both integrated in the same library. Is the plan to eventually migrate to strictly using BaseComponent's descendents?

Tbh I don't like the api of BaseComponent right now and probably will just remove them at some point. Also Im not sure yet if I want to make a separate package for a set of ui components or keep it in the core package.

cybertheory:
Gotcha! Will try and implement. And yea my two cents are if you are keeping them then we should strictly build off the components. If you want to stick to low level dom stuff then spin it off onto a starter package for component libraries. also! I think you should personally double down on basic components like flutter, because as of right now it seems a bit unclear what Jaspr provides in that area since I could probably just do that with Dart:html...

schultek:
I'm going to close this as there hasn't been any activity.

PR #156, opened 2024-01-14
Fontfamilybug

cybertheory:
## Description

Fixed bug with custom font family - Value was stored '"font-family"' instead of "font-family"

## Type of Change

- [ ] ❌ Breaking change
- [ ] ✨ New feature or improvement
- [x] 🛠️ Bug fix
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

## Ready Checklist

### DIFF
--- a/packages/jaspr/lib/src/ui/styles/properties/text.dart
+++ b/packages/jaspr/lib/src/ui/styles/properties/text.dart
@@ -26,7 +26,7 @@ class FontFamily {
   final String value;
 
   /// Constructs a [FontFamily] from a custom font name
-  const FontFamily(String value) : value = '"$value"';
+  const FontFamily(this.value);
 
   const FontFamily._generic(this.value);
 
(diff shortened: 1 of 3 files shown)


schultek:
@cybertheory Please run `melos format` on this and your other prs. You probably have a different line length configured than the 120 that jaspr uses.

packages/jaspr/lib/src/ui/styles/properties/text.dart
```diff
@@ -26,7 +26,7 @@ class FontFamily {
   final String value;
 
   /// Constructs a [FontFamily] from a custom font name
-  const FontFamily(String value) : value = '"$value"';
```
schultek, line 29: Font family names are supposed to be wrapped in double quotes: https://developer.mozilla.org/en-US/docs/Web/CSS/font-family?retiredLocale=de

cybertheory: The problem is it renders as a string within a string in inline styling, so I think it gets sanitized
Screenshot 2024-01-18 at 9 25 50 AM

cybertheory: I know the font name is wrong but even with the correct name it seems to work only without the quotes

cybertheory: https://stackoverflow.com/questions/32462939/inline-custom-css-fonts-in-html5

schultek: Which browser are you using? For me it works also inline, even if its escaped.

cybertheory: arc

schultek: isn't that also chromium based?

cybertheory: yup

schultek:
Fixed this by using single quotes instead of double quotes around the font family name.

PR #158, opened 2024-01-14
fix incorrect named param for "Style"

exoad:
## Description

the name for the "List<StyleRule>" parameter was "rules" which does not align with the code which prompts for a named parameter called "styles"

## Type of Change

- [ ] ❌ Breaking change
- [ ] ✨ New feature or improvement
- [ ] 🛠️ Bug fix
- [ ] 🧹 Code refactor
- [X] 📝 Documentation
- [ ] 🗑️ Chore

## Ready Checklist

### DIFF
--- a/docs/core/styling.mdx
+++ b/docs/core/styling.mdx
@@ -215,7 +215,7 @@ The `Styles` class is great for defining a single set of css properties e.g. to 
 To define a stylesheet with a set of style rules use the `Style` (without *s*) component.
 
 ```dart
-Style(rules: [
+Style(styles: [
     StyleRule(
         selector: const Selector.id('content'),
         styles: Styles.box(


exoad:
see https://github.com/schultek/jaspr/blob/098db73cc7bd28dd0c0d4361e8ce8b8a8a27fa30/packages/jaspr/lib/src/components/style.dart#L6

cybertheory:
I think because it refers to the StyleRule object which is a <style> block, the Styles objects you are referring to are inline styles

exoad:
> I think because it refers to the StyleRule object which is a <style> block, the Styles objects you are referring to are inline styles

Huh, I am referring to the documentation calling one of the parameters in the `Style` class's constructor incorrectly. 

In jaspr, it is:
https://github.com/schultek/jaspr/blob/098db73cc7bd28dd0c0d4361e8ce8b8a8a27fa30/packages/jaspr/lib/src/components/style.dart#L6

While the docs mention the construction of a `Style` object like this:

```dart
Style(rules: [
    StyleRule(
```

But it should be defined as 

```dart
Style(styles: [
```

Sorry for the confusion.

cybertheory:
I get what you are saying, but why would styles be a better name than  rules? keep in mind that you can use MediaRules there as well.

These 'rules' are just the CSS rules within the style block not the actual Style block so it wouldn't make sense to call them 'styles'. 

Personally, I think at a low level calling them styles makes sense and if you want to abstract that away within your own projects you can!

exoad:
> I get what you are saying, but why would styles be a better name than rules? keep in mind that you can use MediaRules there as well.
> 
> These 'rules' are just the CSS rules within the style block not the actual Style block so it wouldn't make sense to call them 'styles'.
> 
> Personally, I think at a low level calling them styles makes sense and if you want to abstract that away within your own projects you can!

I am saying the docs do not reflect the code...

exoad:
It doesn't matter if the docs are called "rules," I am suggesting this PR in order to fix the naming error found in the code in conflict with the naming found in the quickstart guide.

cybertheory:
Oh I see please disregard my previous comments then

schultek:
Thanks.

PR #198, opened 2024-04-02
fix: Forward errors to main isolate in renderHtml

### LINKED ISSUE #129: fix: Jaspr does not properly return a 500 error when an error occurs durring rendering
**Description**

If a widget throws an exception during rendering (for example during `preloadState`) the exception is not bubbled up to shelf, and thus no response is sent to the browser, which looks in Firefox to just be an infinite load.

**Steps To Reproduce**

* Create a Component that throws an exception during `preloadState`.
* Load the page containing the component.

**Doctor Output**

[✓] Jaspr CLI (Version 0.9.2)
  • Dart Version 3.1.3 (stable) (Tue Sep 26 14:25:13 2023 +0000) on "windows_x64" at C:\Users\Jeff Ward\Projects\flutter\bin\cache\dart-sdk\bin\dart.exe
  • Running on windows "Windows 10 Home Single Language" 10.0 (Build 19045) - Locale en-US
  • Analytics: Enabled

[✓] Current Project
  • Dependencies on core packages:
    • jaspr: ^0.9.2
    • jaspr_builder: ^0.9.2 (dev)
    • jaspr_router: ^0.3.0
  • Uses server-side rendering: true
  • Uses experimental compilers: false
  • Uses flutter embedding: false

**Expected Behavior**

Server should properly return a 500 error.

**Additional Context**

This is because the `Isolate.spawn` does not automatically propagate error back to the calling isolate. I have a fix locally, and I'll work on getting a PR up in the next few days if you're open to it.

fuzzybinary:
## Description

Errors when calling `renderHtml` were silently failing and resulting in timeouts because the exception wasn't automatically passed to the spawning isolate. Add an separate `errorPort` to support sending the error to the calling isolate which will properly throw it back the the handler.

This probably also needs to be done for `renderData`.

Partially fixes #129

## Type of Change

🛠️ Bug fix 

## Ready Checklist

### DIFF
--- a/packages/jaspr/lib/src/server/render_functions.dart
+++ b/packages/jaspr/lib/src/server/render_functions.dart
@@ -6,15 +6,20 @@ import 'server_app.dart';
 import 'server_binding.dart';
 
 /// This spawns an isolate for each render, in order to avoid conflicts with static instances and multiple parallel requests
-Future<String> renderHtml(SetupFunction setup, Uri requestUri, Future<String> Function(String) loadFile) async {
+Future<String> renderHtml(SetupFunction setup, Uri requestUri,
+    Future<String> Function(String) loadFile) async {
   if (Jaspr.useIsolates) {
     var port = ReceivePort();
+    var errorPort = ReceivePort();
+    var resultCompleter = Completer<String>.sync();
+    errorPort.listen((message) {
+      if (message is List) {
+        resultCompleter.completeError(message[0]);
+      }
+    });
 
     var message = _RenderMessage(setup, requestUri, port.sendPort);
-
-    await Isolate.spawn(_renderHtml, message);
-
-    var resultCompleter = Completer<String>.sync();
+    await Isolate.spawn(_renderHtml, message, onError: errorPort.sendPort);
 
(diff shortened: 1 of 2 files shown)


packages/jaspr/lib/src/server/render_functions.dart
```diff
@@ -6,15 +6,20 @@ import 'server_app.dart';
...
-Future<String> renderHtml(SetupFunction setup, Uri requestUri, Future<String> Function(String) loadFile) async {
+Future<String> renderHtml(SetupFunction setup, Uri requestUri,
+    Future<String> Function(String) loadFile) async {
   if (Jaspr.useIsolates) {
     var port = ReceivePort();
+    var errorPort = ReceivePort();
+    var resultCompleter = Completer<String>.sync();
+    errorPort.listen((message) {
+      if (message is List) {
+        resultCompleter.completeError(message[0]);
```
schultek, line 17: We should include the stack trace using `StackTrace.fromString(message[1])`

schultek:
Thanks for your contribution. I've merged this and will add the same for the renderData method.

PR #209, opened 2024-04-07
Added main to DOM components

carmas123:
## Description

Changed the generate_html.dart to be able to pass a custom tag value from json definition file, in order to be able to create a "main" html tag usgin a custom function "main_" without have conflicts with the main() Dart method.

## Type of Change

✨ New feature or improvement

## Ready Checklist

- [X] I updated/added relevant documentation (doc comments with `///`).

### DIFF
--- a/packages/jaspr/lib/src/components/html/content.dart
+++ b/packages/jaspr/lib/src/components/html/content.dart
@@ -260,6 +260,26 @@ Component h6(List<Component> children,
   );
 }
 
+/// The &lt;main&gt; HTML element represents the primary content of a document, distinct from content that is repeated across multiple pages such as site headers, footers, and navigation bars.
+Component main_(List<Component> children,
+    {Key? key,
+    String? id,
+    String? classes,
+    Styles? styles,
+    Map<String, String>? attributes,
+    Map<String, EventCallback>? events}) {
+  return DomComponent(
+    tag: 'main',
+    key: key,
+    id: id,
+    classes: classes,
+    styles: styles,
+    attributes: attributes,
+    events: events,
+    children: children,
+  );
+}
+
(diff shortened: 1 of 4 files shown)


packages/jaspr/lib/src/components/html/content.dart
```diff
@@ -260,6 +260,26 @@ Component h6(List<Component> children,
   );
 }
 
+/// The &lt;main&gt; HTML element represents the primary content of a document, distinct from content that is repeated across multiple pages such as site headers, footers, and navigation bars.
+Component main_(List<Component> children,
```
walsha2, line 264: Is `main_` really a better developer experience than:

```dart
DomComponent(
   tag: 'main',
   ...
)
```

Further...

> Note: There must not be more than **one** `<main>` element in a document. 
> https://www.w3schools.com/tags/tag_main.asp

The fact that this DOM element would/should only be used **once in a document**, I really do not think this is a worthwhile addition. Especially due to the conflict with the `main` keyword in dart and the need for the trailing underscore. I bet it adds more confusion than developer benefit.

------

As an end user, I probably would never use this `main_` method. These functions in `content.dart` are simply helper functions to reduce boilerplate, especially for frequently used elements. I dont think **every** HTML tag needs to be here. In fact, many aren't. Directly using `DomComponent` is always available.

carmas123: This is right, as you want there are no problems for me.

schultek: I tried to come up with something better, but I think `main_` is the best we can do.

I don't see a big problem with dx, especially because `main_` is quite well discoverable with IDE autocomplete if you start typing `main` to get the html component.
So I don't think it hurts to have it in the set of html components.

schultek (approved):

PR #250, opened 2024-07-13
fix: add html unescape in hydrated data

siesdart:
## Description

In Jaspr, data is escaped and sent to the client while hydrating. However, there is no unescape process in the client now, so data which contain escapable characters like '&' is different from the server's original on the client. I fixed this by adding unescape progress in the client.

## Type of Change

- 🛠️ Bug fix

## Ready Checklist

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased patch
+
+- Fixed missing html unescape in hydrated data.
+
 ## 0.13.3
 
 - Added support for custom models as parameters to `@client` components.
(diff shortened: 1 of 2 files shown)


schultek:
I indeed missed the unescaping. However this can be a simple string replace operation since the only things that are escaped are 
- `&` to `&amp;`
- `<` to `&lt;`
- `>` to `&gt;`

So something like `str.replaceAll('&lt;', '<').replaceAll('&gt;', '>').replaceAll('&amp;', '&)` or some more optimized version of that.

siesdart:
I modified the code as you said. Does it look better?

schultek (approved):
I changed the order so it doesn't affect itself. but then LGTM.

Thanks for this pr.

PR #256, opened 2024-07-23
feat: activate the content section of tailwind.config.js

siesdart:
## Description

Currently, in `japr_tailwind` package, the `content` section of the `tailwind.config.js` file is completely ignored during the build process, and only the pre-specified dart file path is used for the build. However, in actual use, it may sometimes be necessary to additionally register other files as `content` in addition to the dart file. Therefore, if there is no `tailwind.config.js` file, it automatically uses the pre-specified dart file path as now, but if there is a configuration file, I think it makes more sense to apply the `content` section in the file. Therefore, I modified the code accordingly and suggest this.

## Type of Change

✨ New feature or improvement

## Ready Checklist


packages/jaspr_tailwind/lib/builder.dart
```diff
@@ -46,15 +44,15 @@ class TailwindBuilder implements Builder {
...
         if (options.config.containsKey('tailwindcss')) options.config['tailwindcss'],
         if (hasCustomConfig) ...[
           '--config',
           p.join(Directory.current.path, 'tailwind.config.js').toPosix(),
+        ] else ...[
+          '--content',
+          p.join(Directory.current.path, '{lib,web}', '**', '*.dart').toPosix(true),
         ],
       ],
-      workingDirectory: root,
```
schultek, line 57: This shouldn't be removed. If you think there is a bug or smth let me know.

schultek (changes requested):
Thanks, I think this is a good change.

It should be mentioned in the documentation though that now if you define a custom tailwind config you have to set the content option. Please add this to both the jaspr_tailwind readme and the docs/eco/tailwind docs.

packages/jaspr_tailwind/lib/builder.dart
siesdart: @schultek If Tailwind is not run in the project directory, the `content` paths in `tailwind.config.js` are not detected correctly and CSS is not built properly. Originally, I thought it was common to run the build from the project location when using Tailwind, so may I ask why you set the `workingDirectory` that way?

siesdart:
@schultek As you mentioned, the document has been updated. However, since I am not from an English-speaking country, there is a possibility that there may be awkwardness in the sentences, so I would appreciate it if you could review it. Thank you.

packages/jaspr_tailwind/lib/builder.dart
schultek: This is because build_runner creates a temporary directory where it copies all the project files to and then operates on. So the command should also be run in that temporary directory and not the project dir itself.

If you have a custom config that causes problems can you send me it then I can try to reproduce and see where the problem is.

siesdart: As in the example I added to the documentation, even if you just add `content: ['./{lib,web}/**/*.dart']`, Tailwind won't be able to find the content location properly because the working directory is not the project root directory.

siesdart: Also, after checking with `print`, I found that the temporary folder created by `build_runner` and the path in the `root` variable in the original code before I modified it seem to be different. However, there has been no problem in using it so far. Below is the result confirmed on my computer:

`print(scratchSpace.fileFor(buildStep.inputId).path)`: /private/var/folders/tr/n368kyf54klb2cm37fw2dy_w0000gn/T/scratch_spaceTVDt3X/web/styles.tw.css
`print(root)`: /Users/hyunwoopark/.pub-cache/hosted/pub.dev/jaspr_tailwind-0.2.0

schultek: Ah thanks for debugging this, then the root var is wrong and needs to be set to the scratch space dir.

siesdart: Then, if `tailwindcss` is executed from the scratch space dir as you said, all files belonging to the `content`, such as files under a `lib` folder, should be copied to the same relative path based on the scratch space dir for the `content`. However, `scratchSpace.ensureAssets` copies files under a `lib` folder to a `packages` folder corresponding to its package, not directly under the scratch space dir. Therefore, it is difficult to detect all files belonging to the `content` and copy them to the scratch space dir as they are to build Tailwind, and I don't know why it is necessary to do that. I am curious to know your thoughts on this. Thanks.

schultek: Thanks for diving into this, you changed my mind. It indeed seems to make more sense to run this in the normal project dir. Just the output should still be written to the scratch space.

I will resolve this and do some more testing.

packages/jaspr_tailwind/lib/builder.dart
```diff
@@ -34,8 +34,6 @@ class TailwindBuilder implements Builder {
     var assets = await buildStep.findAssets(Glob('{lib,web}/**.dart')).toList();
     await Future.wait(assets.map((a) => buildStep.canRead(a)));
```
schultek, line 35: When using a custom config with other content files, the tailwind command won't be rerun when those other files change. Do you think this is a problem?

siesdart: Well, it may be a bit cumbersome to use, but since the main files are dart files, I don't think it's a serious problem. If you were to modify this part, you would have to read the config js file and extract the content part in dart. Is it possible to implement this?

schultek: I think its ok for now, but a welcome improvement for a future PR if you want. So I'm going to merge this now and also add a hint that this isn't watching any additional files added to content.

schultek (approved):

schultek:
Thanks again for your work on this, its really appreciated.

Issue #258, opened 2024-07-26
feat: Ability to custom headers on render.


theachoem:
**Description**

Ability to override headers on server before response back to user. I think it could look something like this:

```dart
return serveApp((request, render) async {
  Map<String, String> headers = await setUserSession(request.headers);

  return render(
    headers: headers, <----
    server.Document(
      styles: styles,
      base: '/',
      body: App(),
    ),
  );
})
```

Once headers are passed to `render`, render can take those headers to construct and return a Response.

```dart
return Response(
  200,
  headers: headers,
  body: body,
)
```

**Additional Context**

- Reference from discord:
https://discord.com/channels/993167615587520602/993167616048898059/1266415916225269792

theachoem:
What I requested above could be done through middleware. 

My actual problem is on each page, when I request an API, token can be expired which needed to re-authenticate and set new the token to cookies / headers.

```dart
class HomeView extends StatefulComponent {
  const HomeView({super.key});

  @override
  State<HomeView> createState() => HomeViewState();
}

class HomeViewState extends State<HomeView> with PreloadStateMixin {
  Response? response;

  @override
  Future<void> preloadState() {
    response = await Dio().get('/api/v2/products').catchError((e) {
      if (expired(e)) {
        reauthenticate();
        setCookiesToHeadersBeforeResponseToUser(); <------ Is this possible?
      }
    });
  }

  @override
  Iterable<Component> build(BuildContext context) sync* {
    yield text(response)
  }
}
```

theachoem:
I no longer need this, serverpod cookie works perfectly with jaspr for this.

PR #259, opened 2024-07-26
feat: allow response headers to be edited from context.binding

theachoem:
## Description
Allow server component to set headers via ServerAppBinding. Previously, jaspr return HTML string, but for this PR, I have change to Response instead which will construct headers with content type & body in single place `render` method.

In server component, to override header, just do as following:
```dart
(context.binding as ServerAppBinding).responseHeaders['Your-Custom-Header'] = "This is value";
```

Inspired by Ruby on Rails where you can set response.headers from controller.

## Type of Change

✨ New feature or improvement

## Ready Checklist

## Related issues
- https://github.com/schultek/jaspr/issues/258


theachoem:
I no longer need this, serverpod cookie works perfectly with jaspr for this.

PR #269, opened 2024-08-11
improvement: use package:web

schultek:
## Type of Change

- ❌ Breaking change


simolus3:
@schultek Would a contribution to finish this work be welcome or are you actively working on this?

schultek:
Hi, contributions are always appreciated!

For this one the implementation is fairly finished, I just need to fix the tests and also do some more testing in general.

If you want to help, a review or testing would be amazing.

packages/jaspr/lib/src/foundation/constants.dart
```diff
@@ -71,7 +71,7 @@ const double precisionErrorTolerance = 1e-10;
 /// backed by the same kind of object. Thus a double `0.0` is identical
 /// to an integer `0`. This is not true for Dart code running in AOT or on the
 /// VM.
-const bool kIsWeb = identical(0, 0.0);
+const bool kIsWeb = bool.fromEnvironment('dart.library.js_util');
```
simolus3: Legacy interop libraries are set to be removed, so I think it's better to use `dart.library.js_interop` for this to work in the future.

schultek: True. I just copied this over from flutter without thinking.

schultek: Done

PR #284, opened 2024-08-27
Fix an issue where resolution of webDir would hang.

dam5s:
This solves issues when running the backend server during tests on a Windows machines where the existing solution would infinitely loop in search of the directory.

## Description

Previous implementation would look for the `pubspec.yaml` based on the location of the executable. When running tests on my Windows machine, that location is some random temporary directory. I added a fallback resolution based on the working directory (which is the project directory during tests).

I also made sure we would throw an exception if we could not actually resolve the directory, this informs the user and lets them set an environment variable instead.

The previous implementation would never resolve any directory on Windows machines during tests because the root dir would not satisfy the exit condition of the loop.

## Type of Change

🛠️ Bug fix 

## Ready Checklist

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased patch
+
+- Fixed an issue where webDir would never resolve during tests.
+
 ## 0.15.1
 
 - Include and setup `jaspr_lints` in newly created projects.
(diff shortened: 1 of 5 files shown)


packages/jaspr/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.15.1
```
```suggestion
## Unreleased patch
```

schultek (changes requested):
Thanks for making this, looks good to me.

Adjust the changelog as shown, then I can approve.

dam5s:
> Thanks for making this, looks good to me.
> 
> Adjust the changelog as shown, then I can approve.

Done, thanks!

dam5s:
Just rebased, it should be a non-conflicting single commit now.

schultek (approved):
Thanks

Issue #279, opened 2024-08-22
Global variables


mjablecnik:
Hello, is somehow possible to define css variables for whole html document?

For example this for theming?:
```
/* Dark theme */
.dark-theme {
    --header-bg: #222;
    --header-text: #ddd;
    --nav-bg: #444;
    --nav-text: #ddd;
    --nav-hover-bg: #666;
    --section-bg: #333;
    --footer-bg: #222;
    --footer-text: #ddd;

    --btn-primary-bg: #555;
    --btn-primary-text: #ddd;
    --btn-primary-hover-bg: #666;
    --btn-primary-active-bg: #444;

    --btn-secondary-bg: #007B9E;
    --btn-secondary-text: #ddd;
    --btn-secondary-hover-bg: #008CBA;
    --btn-secondary-active-bg: #005F7A;

    --border-radius: 5px;
}
```

Or should I define my theme into dart variables and change theme by changing some global theme variable?
What is better approach?

mjablecnik:
Hello, so I created very simple InheritedComponent in my project for change themes (light and dark modes):

![Screenshot 2024-08-22 at 17 17 37](https://github.com/user-attachments/assets/a65dc4e5-55e5-4871-b154-21824f1e75b4)

But I discovered that when I want to use it in my project so I need the context but in styles of every component, is context missing because it is static variable:

![Screenshot 2024-08-22 at 17 26 17](https://github.com/user-attachments/assets/de4cfb9b-b8b3-4948-852e-364b1da622a4)

So now I don't know how to change my theme other than by css variables.

schultek:
I suggest using 'data-theme' attribute on the html tag and then using it in a css attribute selector to have separate styles for light and dark.

I will write a more detailed answer when I'm back at the laptop, writing this from the phone rn.

mjablecnik:
@schultek Ok, I will wait for some simple example.

schultek:
So I recommend doing light/dark theming like this:

In `main.dart` (or somewhere else) add the top-level styles using the `@css` annotation:

```dart

@css
final styles = [
  css('body')
      .box(minHeight: 100.vh)
      // Use the background theme value
      .raw({'background-color': 'var(--background-color)'}),
];

// Define the theme values
@css
final List<StyleRule> theme = [
  css('html', [
    css('&[data-theme=\'light\']').raw({
      '--background-color': '#fff',
      '--text-color': '#121416d8',
      // other values...
    }),
[...]

PR #288, opened 2024-09-08
Add variable support

mjablecnik:
## Description

Hello I added support variables for colors and units. https://github.com/schultek/jaspr/issues/279
Right now you can define variables in your html:
```
      .raw({
        '--background-color': '#fff',
        '--text-color': Colors.blue.value,
        '--text-size': 14.pt.value,
        // other values...
      }),

```
and use it as: `Color.variable('--text-color')` or `Unit.variable('--text-size')`

## Type of Change

✨ New feature or improvement

## Ready Checklist

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 0.16.0
-- Added css variable for Unit, Color and Angle.
+- Added css variable for Unit, Color, Angle and FontFamily.
 
 ## 0.15.0
 
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.16.0
+- Added css variable for Unit, Color and Angle.
+
 ## 0.15.0
 
 - Added support for using `@css` and `@encoder`/`@decoder` across other packages.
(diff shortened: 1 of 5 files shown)


schultek:
Nice change. Can you also add this for `Angle`, then it should cover probably >90% of all uses of variables.

And what do you think about calling it `.var()` instead of `.variable()`?

schultek (commented):
Nice change. Can you also add this for Angle, then it should cover probably >90% of all uses of variables.

And what do you think about calling it .var() instead of .variable()?

Dont forget to update the changelog

mjablecnik:
Hello, I added variable also for Angle and updated changelog.
I also wanted to call it as .var() instead of .variable() but there is problem because `var` is Dart key word:

Screenshot 2024-09-09 at 0 09 43
vs
Screenshot 2024-09-09 at 0 09 24
or
Screenshot 2024-09-09 at 0 15 00

So we must name it somehow else then only .var()

packages/jaspr/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.16.0
```
```suggestion
## Unreleased minor
```

mjablecnik:
@schultek Is everything ok? Can you already merge it?

schultek (approved):
LGTM

PR #293, opened 2024-09-11
Fix tests when running with package:web

simolus3:
This adds onto https://github.com/schultek/jaspr/pull/269#issuecomment-2343343221 to fix failing tests.

The main failures essentially boil down to us having to fake an `Event` class when running in the Dart VM that needs to be close to the `Event` class in JS.


schultek:
Looks like my actions are only running for PRs to main. Anyways please run `melos run format`, I see some formatting changes.

schultek (approved):
Approving, so we can continue on the main "web" branch

Issue #299, opened 2024-09-13
Permanently save data


mjablecnik:
Hello, I need save some variable value in my page (for theme and locale) for prevent reset the value after reload of page.
I have saved it in statefulComponent and I would like to somehow permanently save it during update and load after reload page.

I think that I can do this with cookies but I do not know how to save or load it in Jaspr.

Can you give me some example how can I do it?

schultek:
You can either use `document.cookie` for cookies, or `window.localStorage`.

schultek:
For further help I think we continue this better on discord.

mjablecnik:
> You can either use document.cookie for cookies, or window.localStorage.

@schultek Can you show me some example in Jaspr?

PR #301, opened 2024-09-13
Add theme support

mjablecnik:
## Description

Hello, I created simple component for set up theme and also added one simple example how to use it.
I tried to have it very simple but understandable how it works and I hope that everything is clear.
I thought a long of time how to do it and this is my best result..

Let me know what do you think about it and if you approve it I will add also documentation page about theming.

## Type of Change

✨ New feature or improvement

## Ready Checklist


mjablecnik:
Right now theme is not persisted so when you reload page theme reset into default.
For fix it I will need a help with this issue: https://github.com/schultek/jaspr/issues/299

schultek:
I think this is better suited as a package.

I'm still not sure how or even if I want theming be part of the core framework, since there are so many different ways for webdev to style and theme your site, and each one has advantages and disadvantages. 

Also theming is tied a lot to the actual design system you use. In Flutter the Theme is also part of the material library and there isn't a "general" Theme class. 

I'm happy to link any third party package in the docs.

mjablecnik:
Ok, thank you very much. I thought about if it should be a package or part of Jaspr because there is only a few classes. But ok. Now it is clear for me and I am going to create my own package with theme and maybe something else also..

PR #393, opened 2025-03-01
add support for onClick link override and fix Bulma sample

slightfoot:
## Description

I noticed the Bulma CSS example tabs no longer saved their state. This got me looking and a classic issue was found. There was `a` link tag used within the `li` which contains the onClick handler.

https://github.com/schultek/jaspr/blob/52533b98176ff3a51570350b22e0e661d909168a/apps/jaspr_pad/samples/bulma/tabs.dart#L53

This means the link will always reload the page and not let the outer onClick handler take place. It is still correct however to put a link inside the `li` tag as anchors will still action even if javascript is disabled, they also are followed by bots and screen-readers for the visually impared. So moving the onClick handler from the `li` to the `a` tag would be the correct fix. 

This leads to the issue resolved here. The default `events()` function does not allow for preventing the default handler to execute. Which is correct behaviour as their is very little now apart from `a` tags that have the default event behaviour.

This can be resolved a few ways: 
  1. onClick handler can return false (classic html 4)
  2. onClick can call preventDefault on the event given to the callback (html 5)

So this PR adds a `onClick` parameter to the `a` tag function builder that injects the event preventDefault action so the link does not action and cause a browser navigation event. If a Jaspr user does not want this behaviour that can still override the `events` parameter and use the `events()` generator.

https://github.com/slightfoot/jaspr/blob/9871addf7bfcc03eadf948d37ce3a05f4b61409f/packages/jaspr/lib/src/components/html/text.dart#L56

https://github.com/slightfoot/jaspr/blob/9871addf7bfcc03eadf948d37ce3a05f4b61409f/apps/jaspr_pad/samples/bulma/tabs.dart#L56

> [!WARNING]
> This is untested.

## Ready Checklist

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Unreleased
+## Unreleased patch
 
 - Added `onClick` override to `a` html tag. Fixing issues with default action handling. @slightfoot
 
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Added `onClick` override to `a` html tag. Fixing issues with default action handling. @slightfoot
+
 ## 0.18.0
 
 - **BREAKING** Changed `AppBinding`s `Uri get currentUri` to `String get currentUrl`.
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Unreleased patch
 
-- Added `onClick` override to `a` html tag. Fixing issues with default action handling. @slightfoot
+- Added `onClick` override to `a` html tag. Fixing issues with default action handling.
 
 ## 0.18.0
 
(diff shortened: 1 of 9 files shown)


packages/jaspr/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
```
```suggestion
## Unreleased patch
```

packages/jaspr/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+- Added `onClick` override to `a` html tag. Fixing issues with default action handling. @slightfoot
```
schultek: I never add contributor tags to the changelog, I feel it would be unfair to other contributors to start now.

packages/jaspr/lib/src/components/html/text.dart
```diff
@@ -47,7 +50,14 @@ Component a(List<Component> children,
       if (type != null) 'type': type,
       if (referrerPolicy != null) 'referrerpolicy': referrerPolicy.value,
     },
-    events: events,
+    events: {
```
schultek, line 52: This file is auto-generated by running 'melos run generate:html'.
Script is at 'packages/jaspr/tool/generate_html.dart' and data source '.../tool/data/html.json'.
So we cannot just modify it, but need to make a solution in the script.

schultek (changes requested):
Generally I like this very much, I had this same problem many times but never thought of just defaulting to doing preventDefault in a custom handler.

Though as you see in my comment below the solution cannot directly modify the text.dart file as it is generated.

packages/jaspr/lib/src/components/html/text.dart
schultek: @slightfoot I pushed a change, lmk what you think.

slightfoot: I see the change, that works. reverting these changes

schultek (dismissed):
LGTM

PR #397, opened 2025-03-07
Add --extra-js-compiler-option/--extra-wasm-compiler-option

mkustermann:
This allows jaspr users to pass any arguments to the
  * `dart compile js`
  * `dart compile wasm` compilers.

For example if one wants an `-O4` optimized build but symbols for profiling one may run

  * `jaspr build -O4 --extra-js-compiler-option=--no-minify`
  * `jaspr build -O4 --experimental-wasm --extra-wasm-compiler-option=--no-strip-wasm`

✨ New feature or improvement


schultek (approved):
Nice, LGTM

mkustermann:
Thanks for the quick merge, @schultek :rocket:

PR #399, opened 2025-03-14
Some performance optimizations in the jaspr framework

mkustermann:
* Avoid repeatedly accessing the same `target.type` JS interop field
* Avoid unnecessary map allocation in map literal
* Avoid unnecessary closure allocation in joining maps/strings
* Avoid unnecessary map allocations in joining maps and one is empty

### DIFF
--- a/packages/jaspr/lib/src/framework/dom_component.dart
+++ b/packages/jaspr/lib/src/framework/dom_component.dart
@@ -79,10 +79,10 @@ class DomElement extends ProxyRenderObjectElement {
       renderObject.updateElement(
         component.tag,
         component.id ?? wrappingComponent.id,
-        _join(wrappingComponent.classes, component.classes, (a, b) => '$a $b'),
-        _join(wrappingComponent.styles?.properties, component.styles?.properties, (a, b) => {...a, ...b}),
-        _join(wrappingComponent.attributes, component.attributes, (a, b) => {...a, ...b}),
-        _join(wrappingComponent.events, component.events, (a, b) => {...a, ...b}),
+        _joinString(wrappingComponent.classes, component.classes),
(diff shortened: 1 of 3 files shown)


schultek (approved):
LGTM!

Thanks a lot, great stuff.

packages/jaspr/lib/src/framework/dom_component.dart
```diff
@@ -98,8 +98,18 @@ class DomElement extends ProxyRenderObjectElement {
     );
   }
 
-  T? _join<T>(T? a, T? b, T Function(T a, T b) joiner) {
-    return a != null && b != null ? joiner(a, b) : a ?? b;
+  static String? _joinString(String? a, String? b) {
+    if (a == null) return b;
+    if (b == null) return null;
```
kevmoo, line 103: Not `return a;`?

packages/jaspr/lib/src/framework/dom_component.dart
```diff
@@ -98,8 +98,18 @@ class DomElement extends ProxyRenderObjectElement {
...
+    return '$a $b';
+  }
+
+  static Map<K, V>? _joinMap<K, V>(Map<K, V>? a, Map<K, V>? b) {
+    if (a == null) return b;
+    if (b == null) return null;
```
kevmoo, line 109: Not `return a;`?

packages/jaspr/lib/src/framework/dom_component.dart
mkustermann: Nice catch. Though it was already fixed in b6a8a56c5

packages/jaspr/lib/src/framework/dom_component.dart
mkustermann: Nice catch. Though it was already fixed in b6a8a56c5

PR #434, opened 2025-06-07
Fix: CTA in copy button to be same as text itself

Meldiron:
## Description

Fixes installation command. I clicked copy button, and pasted into CLI. Then I went to get started docs, scrolled down to CLI setup, and ran `jaspr create my_website`, which didnt work.

That's when I noticed command in docs is different than one I ran. I compared with homepage, and it was same. I as confused where I got the wrong command, but when looking through source-code, I noticed wrong command is placed in copy action of the command on homepage.

I checked rest of repository too, and this was the only place with wrong command

## Type of Change

📝 Documentation

## Ready Checklist


schultek (approved):
Oh wow, great catch thanks

PR #447, opened 2025-06-19
[content] Require Dart 3.8, format, and update lints

parlough:
To make contributing a bit easier and more consistent, update constraint to Dart 3.8 with the recent format fixes, set `trailing_commas: preserve`, run the formatter, and update the lints package.


schultek:
When I update min sdk version, I will do it for all packages to stay in sync. But I'm not sure I want to update yet.

PR #458, opened 2025-06-25
[content] Avoid use of `dynamic` and implicit casts

parlough:


packages/jaspr_content/lib/src/page.dart
```diff
@@ -130,6 +132,14 @@ class Page {
   }
 }
 
+extension PageNamespacedData on Page {
+  /// The data specific to this page, namespaced under 'page' in [data].
+  Map<String, Object?> get namespacedPageData => switch (data['page']) {
```
schultek: I think its good to have a more type-safe way to access this, but I don't really like the naming (mainly that its so long).

How about an alternative approach:

`Page.data` will be made a `PageDataMap` that implements normal `Map<String, Object?>` but also gets a typed `page` getter. Then this can be used as `page.data.page` but also still `page.data['site']`.

parlough: I really love that idea! It also allows downstream developers to add extension methods to access their own data to that type specifically rather than all maps.

I took your suggestion and opened a proposal for it in https://github.com/schultek/jaspr/pull/471. Let me know what you think :D

parlough:
@schultek Any concerns with landing this stricter analysis config for Jaspr Content? Didn't require too many changes to the code, but helps avoid potential issues in the future.

Note that the analysis failure looks unrelated.

schultek:
No concerns. Though I prefer to have the same analysis settings for all packages, but doing so for all would be a lot of work.

parlough:
> Though I prefer to have the same analysis settings for all packages, but doing so for all would be a lot of work.

I'm happy to help iteratively get there, landing stricter analysis for a package (or related set) in a PR at a time.

parlough:
@schultek Now that we have the shared analysis config, I think it's a good time to start landing this cleanup. I can land the updates to packages individually then eventually move to the shared config.

PR #471, opened 2025-07-02
[content] Wrap page.data in an extension type with typed getters

parlough:
Extracted from https://github.com/schultek/jaspr/pull/458 in response to the suggestion in https://github.com/schultek/jaspr/pull/458#discussion_r2178456134.

I included `page.data.site` as well as it seems some built-in tooling also relies on that key, but if you don't want to reserve `'site', I'm happy to review it :)

### DIFF
--- a/packages/jaspr_content/lib/src/page.dart
+++ b/packages/jaspr_content/lib/src/page.dart
@@ -59,14 +59,22 @@ class Page {
   /// The route loader that created this page.
   final RouteLoader loader;
 
-  /// Applies changes to the page content or data.
+  /// Applies changes to the page [content] or [data].
+  ///
+  /// If [data] isn't `null` and [mergeData] is `true`,
+  /// values from [data] override those from [Page.data].
+  /// Maps with string keys are deeply merged,
+  /// also favoring values from [data].
+  ///
+  /// If [data] isn't `null` and [mergeData] is `false`,
+  /// [data] replaces the page's entire [Page.data].
   void apply({String? content, Map<String, Object?>? data, bool mergeData = true}) {
     if (content != null) {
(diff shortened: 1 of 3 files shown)


packages/jaspr_content/lib/src/page.dart
```diff
@@ -355,14 +371,33 @@ class _InheritedPage extends InheritedComponent {
...
+  /// Data declared under the 'site' key in [data],
+  /// often loaded from a `_data/site.yml` file and shared across the site.
+  Map<String, Object?> get site => switch (_data['site']) {
+        final Map<String, Object?> siteData => siteData,
+        _ => const {},
+      };
+}
+
+extension DataMergeExtension on Map<String, Object?> {
+  Map<String, Object?> merge(Map<String, Object?> other) {
```
schultek, line 409: We could make this now directly on PageDataMap, then its no longer 'polluting' the Map type

schultek: Also returning a `PageDataMap` directly

schultek:
In general, I really like the idea of using an extension type here.

packages/jaspr_content/lib/src/page.dart
parlough: This ends up being a bit awkward as it also deeply merges the maps which aren't really `PageDataMap`.

I went a different direction and made the extension method private as for the purposes of `Page`, it's more of an implementation detail of the `apply` method. This isn't a breaking change either as the extension was explicitly hidden when exporting the library anyway.

While doing this, I took the opportunity to expand the doc comment on `apply` with more details about how data is merged.

schultek (approved):

PR #481, opened 2025-07-07
[content] fix: subfolders in `Content` are not recognized on Windows

tetsuhou:
## Description

jaspr_content doesn't recognize subfolders in the Windows path format, so a cross-platform library `path` was added to parse the format.

Using `final segments = path.replaceAll(r'\', '/').split('/');` directly is also a way to do it, but I can't evaluate which is better myself.

I'm not good at English, so please let me know if I got something wrong.

## Type of Change

🛠️ Bug fix

## Ready Checklist

### DIFF
--- a/packages/jaspr_content/lib/src/route_loader/route_loader.dart
+++ b/packages/jaspr_content/lib/src/route_loader/route_loader.dart
@@ -8,6 +8,7 @@ import 'dart:collection';
 
 import 'package:jaspr/server.dart';
 import 'package:jaspr_router/jaspr_router.dart';
+import 'package:path/path.dart' as pkg_path;
 
 import '../page.dart';
 import '../secondary_output/secondary_output.dart';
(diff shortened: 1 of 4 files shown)


schultek (approved):
Good catch, thanks.

packages/jaspr_content/lib/src/route_loader/route_loader.dart
```diff
@@ -185,7 +186,7 @@ final indexRegex = RegExp(r'index\.[^/]*$');
 
 abstract class PageSource {
   PageSource(this.path, this.loader, {bool keepSuffix = false}) {
-    final segments = path.split('/');
+    final segments = pkg_path.split(path);
```
schultek: This needs a way to override the path context during tests. My proposal is to add a optional `{pkg_path.Context? context}` parameter to `PageSource`, and then inside `FileSystemLoader` use `fileSystem.path` for this param.


```suggestion
PageSource(this.path, this.loader, {bool keepSuffix = false, pkg_path.Context? context}) {
    final segments = (context ?? pkg_path.context).split(path);
```

schultek (approved):

PR #525, opened 2025-08-19
[jaspr_builder] Update to analyzer v8

parlough:


schultek:
This is blocked by jaspr_lints depending on analyzer 7.x because of custom_lint_builder, which probably won't get any more updates because of the new dart analyzer_plugin work going on. Therefore this will be blocked until I migrate jaspr_lints to the new analyzer_plugin system.

schultek:
`custom_lint_builder` now supports analyzer 8, so I should be able to update as well. That's nice then I can still include it in 0.21.

schultek:
This will be done in #555

PR #530, opened 2025-08-28
Add Filter and BackdropFilter styles properties

ShubhamVG:
## Description
- Added `Filter` class for `filter` and `backdropFilter` css properties in a new file filter.dart.
- Changed implicit type of dynamic to explicit Angle on the add operator of Angle.
- Added the changes to docs and changelog.

## Type of Change

- 🧹 Code refactor
- 📝 Documentation

## Ready Checklist


shubhamlav:
~~Working on it.~~ Done.

shubhamlav:
@schultek - can you give some feedback? I have other CSS properties that I'd like to implement and they might follow a similar pattern of abstract classes and static attributes. I'd like to know if I did it correctly this time or if there are changes to be made. Thanks.

schultek:
Yes will take a look until tomorrow

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
```diff
@@ -0,0 +1,188 @@
+import 'angle.dart';
+import 'color.dart';
+import 'unit.dart';
+
+typedef BackdropFilter = Filter;
```
schultek: I would remove this and just use `Filter` everywhere

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
```diff
@@ -0,0 +1,188 @@
...
+  static const Filter none = Filter._('none');
+
+  // Global values
+  static const Filter inherit = Filter._('inherit');
+  static const Filter initial = Filter._('initial');
+  static const Filter revert = Filter._('revert');
+  static const Filter revertLayer = Filter._('revert-layer');
+  static const Filter unset = Filter._('unset');
+
+  const factory Filter(List<FilterFn> filterValues) = _Filter;
```
schultek: Make this named as `Filter.all` or `.list` so its more explicit that this combines multiple filters.

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
```diff
@@ -0,0 +1,188 @@
+import 'angle.dart';
+import 'color.dart';
+import 'unit.dart';
+
+typedef BackdropFilter = Filter;
+
+class Filter {
```
schultek, line 5: I would flatten the FilterFn class directly into the Filter class. Because currently, the most common case of applying a single filter is a bit verbose:

```dart
filter: Filter([
  FilterFn.blur(8.px),
]),
```

but it should just be

```dart
filter: Filter.blur(8.px)
```

So make all filter functions like `const factory FilterFn.blur([Unit? length]) = _BlurFilter;` directly on the `Filter` class and remove `FilterFn`.

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
```diff
@@ -0,0 +1,188 @@
...
+  const _BlurFilter([this._length]);
+
+  final Unit? _length;
+
+  /// The css value
+  @override
+  String get value => "blur(${_length?.value ?? ''})";
+}
+
+class _BrightnessFilter implements FilterFn {
```
schultek: You can add a shared superclass for these that are all some percentage in filter function.

e.g.
```dart
class _FilterWithPercentage implements Filter {
  _FilterWithPercentage(this._percentage, this._name);

  final double? _percentage;
  final String _name;

  @override
  String get value => "$_name(${_percentage?.numstr ?? ''})";
}
```

and then

```dart
class _BrightnessFilter extends _FilterWithPercentage {
  const _BrightnessFilter(super._percentage) : super('brightness');
}
```

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
```diff
@@ -0,0 +1,188 @@
...
+
+  final List<FilterFn> _filterValues;
+
+  /// The css value
+  @override
+  String get value => _filterValues.map((e) => e.value).join(' ');
+}
+
+abstract class FilterFn {
+  const factory FilterFn.blur([Unit? length]) = _BlurFilter;
```
schultek: I would make all of the params non-nullable, but default to zero, e.g.

`Filter.blur([Unit length = Unit.zero])`
`Filter.brightness([double percentage = 0])`

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
```diff
@@ -0,0 +1,188 @@
...
+    Color? color,
+  }) = _DropShadowFilter;
+  const factory FilterFn.grayscale([double? percentage]) = _GrayscaleFilter;
+  const factory FilterFn.hueRotate([Angle? angle]) = _HueRotateFilter;
+  const factory FilterFn.invert([double? percentage]) = _InvertFilter;
+  const factory FilterFn.opacity([double? percentage]) = _OpacityFilter;
+  const factory FilterFn.sepia([double? percentage]) = _SepiaFilter;
+  const factory FilterFn.saturate([double? percentage]) = _SaturateFilter;
+  const factory FilterFn.url(String url) = _UrlFilter;
+  const factory FilterFn.src(String url) = _SrcFilter;
```
schultek: I cannot find any documentation on this filter?

schultek (commented):
Generally I very much welcome this contribution, very nice addition. I left some comments on the specific api and implementation.

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
shubhamlav: Won't this allow something like this?
```dart
Filter.all([
  Filter.all([]),
  Filter.none,
]);
```

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
shubhamlav: I did think about doing this however there are some properties which defaults too 100% instead of 0. Here is one example from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/sepia#examples_of_correct_values_for_sepia):
```css
sepia(0)     /* No effect */
sepia(.65)   /* 65% sepia */

sepia()      /* Completely sepia */
sepia(100%)
```

We can have a default value for each one but I'm not sure if that would be a good idea in case there are browsers which defaults to different values, and/or this may create ambiguity amongst devs.
The former sounds unlikely but I've never used more than two browsers ever so I'm not sure how browser support works. You may know more about this than me so I need your call on this: whether we keep a default value (different for each filter) or we keep values nullable.

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
shubhamlav: I presume you are referring to url or src here. I actually forgot to mention, there aren't really "filter-function" but the formal syntax on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/filter#formal_syntax) is:
```
filter = 
  none                 |
  <filter-value-list>  

<filter-value-list> = 
  [ <filter-function> | <url> ]+
  
<url> = 
  <url()>  |
  <src()>

<url()> = 
  url( <string> <url-modifier>* )  |
  <url-token>                      

<src()> = 
  src( <string> <url-modifier>* ) 
  ```
  There is an [example](https://developer.mozilla.org/en-US/docs/Web/CSS/filter#try_it) of `url` on the MDN's try-out section, however, there is none for `src` so I have no idea how it works. I only added it since it was present in the formal syntax.

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
shubhamlav: I think one easy fix to this would be to create a `_FilterListVal implements Filter` class and all the list values would extend from it and in `_FilterList`'s `value` method, we could throw an error if a value was detected which is not a subclass of `_FilterListVal`. How does this sound? @schultek

schultek: Yes sounds good, you can add an assert for it. I've done something similar for `Selector` inside `_ChainSelector`.

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
schultek: Ah good catch. I think its ok to then use the different default values, I'm pretty sure these are the same across all browsers. 

From mdn (feel free to double check):
- blur: 0
- brightness: 1
- contrast: 1
- grayscale: 1
- invert: 1
- opacity: 1
- saturate: 1
- sepia: 1

packages/jaspr/lib/src/foundation/styles/properties/filter.dart
schultek: Yes I meant the `src()`. I would remove it since its nowhere mentioned except in the syntax, so maybe its legacy or something.

shubhamlav:
@schultek - I have added some assertions and throwing strings if they fail. Let me know if this error handling is okay or I should instead throw ArgumentError, or any sort of changes that you require.

PR #531, opened 2025-08-28
[jaspr_cli] Work around --define issue in Dart 3.10 dev builds

parlough:
Add a workaround for https://github.com/dart-lang/sdk/issues/61373 so `jaspr build` and `jaspr serve` work as expected even with the issue in 3.10 beta and dev builds. Once resolved, we can revert the workaround, but this way of using `dart` is also supported, so there's no rush to do so.


parlough:
@schultek This fixes the issue blocking me recently since our site is usually developed and built with beta releases of Dart.

However, with this fix, the single child change is working great!! Thank you :D

PR #534, opened 2025-08-30
Share a single base analysis config across workspace packages

parlough:


schultek:
I'm not sure I like this. 

Is it ok for the packages to use the internal package for analysis config, but don't depend on it? Will there be a problem when publishing the packages since they use the internal package (which isn't published)?

It it doesn't need to depend on it for the analysis yaml, can't this simple be one at the root `jaspr_workspace` package?

parlough:
Thanks for taking a look @schultek! I updated the PR description with some more details, fixed the missing dependencies, and responded to your questions here:

> Is it ok for the packages to use the internal package for analysis config, but don't depend on it? Will there be a problem when publishing the packages since they use the internal package (which isn't published)?

I meant to replace the dev-dependency on `package:lints` with one on `package:jaspr_repo`. That way we only need to update the `package:lints` dependency in one place. In terms of publishing, there won't be any problems as these are dev dependencies, which are not considered or included during publishing. 

> It it doesn't need to depend on it for the analysis yaml, can't this simple be one at the root jaspr_workspace package?

Depending on your preferences, I believe there are a few options. Let me know your thoughts :)

1. This solution, with a new internal package that is used as a dev dependency with a `analysis_options.yaml` file in its `lib` directory.
2. Place the `analysis_options.yaml` file in the root directory and include it with a relative path in each package that needs to customize the config.
   - I'm not a huge fan of this as the include path will be different depending on how nested a package is.
3. Add a `lib` directory to the root of the workspace and include the `analysis_options.yaml` file in there.
   - This is fine, but it seems the consensus is that workspace root packages should not contain libraries themselves, but rather be a host to the workspace. Plus I feel it's nice to keep the root directory cleaner.

parlough:
Any new thoughts on this @schultek? This will help me to start unifying towards a shared, stricter set of analysis options.

schultek:
I will take another look once the release is all done and settled

schultek (approved):

PR #540, opened 2025-09-02
[jaspr_pad] Dependency updates and clean up

parlough:
- Removes `yaml_writer` as a dev dependency as it wasn't used anymore.
- Removes `package:js` as a dev dependency and replaces it with `dart:js_interop`
- Tightens a few other dependency constraints


parlough:
I added `/apps` and `/examples` as directories to trigger the analysis workflows now that they're part of the workspace.

PR #552, opened 2025-09-06
feat: Add `All`, `Appearance` and `AspectRatio`  CSS properties.

ShubhamVG:
## Description
This PR adds `All`, `Appearance` and `AspectRatio`  properties along with the updated tests and docs.
I added all the files in box.dart. I figured that is the most appropriate place to put all three styles at. LMK anything needs to be moved anywhere else.
Take a special look at `AspectRatio.auto`. That API may need a change since it is a factory instead of a static variable.
I made a new group test for `AspectRatio` for coverage reasons. I was not sure if this would go to something like unit_test.dart or not.

## Type of Change

- 📝 Documentation

## Ready Checklist

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.21.0-dev.1
 
 - Fixed update check for pre-release versions.
+- Added `All`, `Appearance` and `AspectRatio` styles properties.
 
 ## 0.21.0-dev.0
 
--- a/packages/jaspr/lib/src/foundation/styles/styles.dart
+++ b/packages/jaspr/lib/src/foundation/styles/styles.dart
@@ -39,6 +39,7 @@ abstract class Styles with StylesMixin<Styles> {
     Transition? transition,
     Transform? transform,
     All? all,
+    Appearance? appearance,
     // Flexbox Styles
     FlexDirection? flexDirection,
     FlexWrap? flexWrap,
@@ -128,6 +129,7 @@ abstract mixin class StylesMixin<T> {
     Transition? transition,
     Transform? transform,
     All? all,
+    Appearance? appearance,
     // Flexbox Styles
     FlexDirection? flexDirection,
     FlexWrap? flexWrap,
@@ -199,6 +201,7 @@ abstract mixin class StylesMixin<T> {
         opacity: opacity,
         transform: transform,
         all: all,
+        appearance: appearance,
         shadow: shadow,
         filter: filter,
         backdropFilter: backdropFilter,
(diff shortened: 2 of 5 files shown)


shubhamlav:
Good to see the analyze code test pass :joy: 
It has been failing for the past few weeks since I didn't know I have to use page width 120 :p

schultek:
I would change the order of the parameters as follows:

all: This is actually important, because it changes how it works whether it's applied first or last (or between) other properties. Putting it last would reset all properties when used, which could lead to frustration. I think this should be first of all properties, then it can reset all existing styling of an element, but still allows to then define other properties again.

appearance: I would put it after overflow, since its a general visual property.

aspectRatio: after maxHeight since its a sizing property

schultek:
For the AspectRatio.auto api, I think we should have a const var .auto, which probably covers 90% of usage, and then a .autoOrRatio() factory for the fallback case.

packages/jaspr/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 0.21.0-dev.1
 
 - Fixed update check for pre-release versions.
+- Added `All`, `Appearance` and `AspectRatio` styles properties.
```
schultek: Add a new `## Unreleased 0.21.0-dev.2` header, the .1 is already published

shubhamlav:
@schultek - Done. Anything else?

packages/jaspr/CHANGELOG.md
shubhamlav: Done.

shubhamlav:
Oh, nice to see the coverage report!

packages/jaspr/CHANGELOG.md
schultek: The "Unreleased" is still needed, otherwise the ci won't pick it up

shubhamlav: Right, somehow I didn't see "Unreleased" the first time, only saw the .2 :p

packages/jaspr/lib/src/foundation/styles/styles.dart
```diff
@@ -458,6 +473,9 @@ class _Styles extends Styles {
           '-webkit-user-select': userSelect!.value,
         },
         if (pointerEvents != null) 'pointer-events': pointerEvents!.value,
+        if (all != null) 'all': all!.value,
+        if (appearance != null) 'appearance': appearance!.value,
+        if (aspectRatio != null) 'aspect-ratio': aspectRatio!.value,
```
schultek: Still need to update the position here for the properties.

shubhamlav: Done, moved all 3.

shubhamlav: Should I add a test for that somewhere? To check if 'all' is on the top?

schultek: I just noticed there are a lot other props that aren't where they should be, I will fix it and add a test

schultek (dismissed):

PR #564, opened 2025-09-16
Bug fixing and chores

### LINKED ISSUE #539: fix: css properties returns empty string when takes an empty list
**Description**

While going through some CSS implementations, I realized that there are some e.g., `GridTracks` which takes in a list as argument. Their (css) value function is as follows:
```dart
String get value => tracks.map((t) => t.value).join(' ');
```
However, no where in the implementation is an assertion or anything to check whether the list is empty in which case, it should either crash or return a default value.

That means that this:
```dart
css('#grid').styles(
  display: Display.grid,
  gridTemplate: GridTemplate(columns: GridTracks([])),
),
```
gets converted to:
```css
#grid {
  display: grid;
  grid-template-columns: ; /* value is empty, problematic */
}
```

My question is: what to we do/go for? Do we throw an error, return a default value, or something else?
Even though these results can only be achieved by using the API very incorrectly, I still don't think it would be wise to leave it as it is.
Actually, in my previous PR where I implemented `Filter`, I had an assertion which throws `Filter.list cannot be empty.` in such a case. I don't know why didn't I instead made it return the none value (the default according to MDN) nor do I know if that would have been a good idea. I want your feedback & suggestion. BTW, I can work on this issue myself. Just make your call and assign this to me.

**Steps To Reproduce**

1. Create a Jaspr project and make a div. In its styling, give it display: grid and gridTemplate: with empty tracks.
2. jaspr serve it and open it in browser.
3. Find the element, click on it and check its CSS properties. The grid-template-thing's value would be empty along with a caution/error beside it.

ShubhamVG:
Closes #539
## Description
- Closes #539: Added assertions and throw string messages whenever a possible API misuse could happen.
- Replaced a broken link for the docs website.  

## Type of Change

- ❌ Breaking change

- 🛠️ Bug fix

- 🗑️ Chore

## Ready Checklist

- [ ] Write tests

### DIFF
--- a/packages/jaspr/lib/src/foundation/styles/properties/box.dart
+++ b/packages/jaspr/lib/src/foundation/styles/properties/box.dart
@@ -453,7 +453,19 @@ enum BoxSizing {
   const BoxSizing(this.value);
 }
 
-abstract class BoxShadow {
+class BoxShadow {
+  const BoxShadow._(this.value);
+
+  final String value;
+
+  // Keyword and global values
+  static const BoxShadow none = BoxShadow._('none');
+  static const BoxShadow initial = BoxShadow._('initial');
+  static const BoxShadow inherit = BoxShadow._('inherit');
+  static const BoxShadow revert = BoxShadow._('revert');
+  static const BoxShadow revertLayer = BoxShadow._('revert-layer');
+  static const BoxShadow unset = BoxShadow._('unset');
+
   const factory BoxShadow({required Unit offsetX, required Unit offsetY, Unit? blur, Unit? spread, Color? color}) =
       _BoxShadow;
 
@@ -466,11 +478,11 @@ abstract class BoxShadow {
   }) = _InsetBoxShadow;
 
   const factory BoxShadow.combine(List<BoxShadow> shadows) = _CombineBoxShadow;
-
-  String get value;
 }
 
-class _BoxShadow implements BoxShadow {
+abstract class _ListableBoxShadow implements BoxShadow {}
+
+class _BoxShadow implements _ListableBoxShadow {
   const _BoxShadow({
     required this.offsetX,
     required this.offsetY,
@@ -508,8 +520,29 @@ class _CombineBoxShadow implements BoxShadow {
 
   final List<BoxShadow> shadows;
 
+  bool _shadowsListable() {
+    if (shadows.isEmpty) {
+      throw '[BoxShadow.combine] cannot be empty.';
+    }
+
+    for (final shadow in shadows) {
+      if (shadow is _CombineBoxShadow) {
+        throw 'Cannot nest [BoxShadow.combine] inside [BoxShadow.combine]';
+      }
+
+      if (shadow is _ListableBoxShadow) {
+        throw 'Cannot use ${shadow.value} as a list item, only standalone use supported.';
+      }
+    }
+
+    return true;
+  }
+
   @override
-  String get value => shadows.map((s) => s.value).join(', ');
+  String get value {
+    assert(_shadowsListable());
+    return shadows.map((s) => s.value).join(', ');
+  }
 }
 
 class Cursor {
--- a/packages/jaspr/lib/src/foundation/styles/properties/box.dart
+++ b/packages/jaspr/lib/src/foundation/styles/properties/box.dart
@@ -61,8 +61,8 @@ class __Border implements Border {
 
   const __Border({this.style = BorderStyle.solid, this.color, this.width});
 
-  bool _propertyStrNotEmpty(String property) {
-    if (property.isEmpty) {
+  bool _attributesNotEntirelyNull() {
+    if (style == null && color == null && width == null) {
       throw '[Border] cannot have all attributes null.';
     }
     return true;
@@ -70,13 +70,14 @@ class __Border implements Border {
 
   @override
   Map<String, String> get styles {
-    final borderProperty = <String>[
-      if (style != null) style!.value,
-      if (color != null) color!.value,
-      if (width != null) width!.value,
-    ].join(' ');
-    assert(_propertyStrNotEmpty(borderProperty));
-    return <String, String>{'border': borderProperty};
+    assert(_attributesNotEntirelyNull());
+    return <String, String>{
+      'border': [
+        if (style != null) style!.value,
+        if (color != null) color!.value,
+        if (width != null) width!.value,
+      ].join(' '),
+    };
   }
 }
 
--- a/packages/jaspr/lib/src/foundation/styles/properties/box.dart
+++ b/packages/jaspr/lib/src/foundation/styles/properties/box.dart
@@ -61,14 +61,23 @@ class __Border implements Border {
 
   const __Border({this.style = BorderStyle.solid, this.color, this.width});
 
+  bool _propertyStrNotEmpty(String property) {
+    if (property.isEmpty) {
+      throw '[Border] cannot have all attributes null.';
+    }
+    return true;
+  }
+
   @override
-  Map<String, String> get styles => {
-    'border': [
+  Map<String, String> get styles {
+    final borderProperty = <String>[
       if (style != null) style!.value,
       if (color != null) color!.value,
       if (width != null) width!.value,
-    ].join(' '),
-  };
+    ].join(' ');
+    assert(_propertyStrNotEmpty(borderProperty));
+    return <String, String>{'border': borderProperty};
+  }
 }
 
 class _OnlyBorder implements Border {
--- a/packages/jaspr/lib/src/foundation/styles/properties/box.dart
+++ b/packages/jaspr/lib/src/foundation/styles/properties/box.dart
@@ -531,7 +531,7 @@ class _CombineBoxShadow implements BoxShadow {
         throw 'Cannot nest [BoxShadow.combine] inside [BoxShadow.combine]';
       }
 
-      if (shadow is _ListableBoxShadow) {
+      if (shadow is! _ListableBoxShadow) {
         throw 'Cannot use ${shadow.value} as a list item, only standalone use supported.';
       }
     }
--- a/packages/jaspr/lib/src/foundation/styles/properties/grid.dart
+++ b/packages/jaspr/lib/src/foundation/styles/properties/grid.dart
@@ -35,8 +35,18 @@ class _GridTracks implements GridTracks {
 
   final List<GridTrack> tracks;
 
+  bool _validGridTracks() {
+    if (tracks.isEmpty) {
+      throw '[GridTracks] cannot be empty.';
+    }
+    return true;
+  }
+
   @override
-  String get value => tracks.map((t) => t.value).join(' ');
+  String get value {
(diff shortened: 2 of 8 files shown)


packages/jaspr/lib/src/foundation/styles/properties/box.dart
```diff
@@ -61,14 +61,24 @@ class __Border implements Border {
 
   const __Border({this.style = BorderStyle.solid, this.color, this.width});
 
+  bool _attributesNotEntirelyNull() {
```
schultek: This can actually be an assert on the constructor, since its simple enough to be const.

`assert(style != null || color != null || width != null, 'At least one of style, color or width must be not null. For no border, use Border.none')`

packages/jaspr/lib/src/foundation/styles/properties/box.dart
```diff
@@ -499,8 +521,29 @@ class _CombineBoxShadow implements BoxShadow {
...
   final List<BoxShadow> shadows;
 
+  bool _shadowsListable() {
+    if (shadows.isEmpty) {
+      throw '[BoxShadow.combine] cannot be empty.';
+    }
+
+    for (final shadow in shadows) {
+      if (shadow is _CombineBoxShadow) {
+        throw 'Cannot nest [BoxShadow.combine] inside [BoxShadow.combine]';
```
schultek: I would actually allow that, there shouldn't be a problem with the css.

packages/jaspr/lib/src/foundation/styles/properties/grid.dart
```diff
@@ -35,16 +35,36 @@ class _GridTracks implements GridTracks {
 
   final List<GridTrack> tracks;
 
+  bool _validGridTracks() {
```
schultek: This can also be a constant assert on the constructor with `tracks.length > 0`

packages/jaspr/lib/src/foundation/styles/properties/grid.dart
```diff
@@ -35,16 +35,36 @@ class _GridTracks implements GridTracks {
...
 
 class GridAreas {
   const GridAreas(this.lines);
 
   final List<String> lines;
 
-  String get value => lines.map((l) => '"$l"').join(kDebugMode ? '\n' : ' ');
+  bool _validGridAreas() {
+    if (lines.isEmpty) {
+      throw '[GridAreas] cannot be empty.';
```
schultek: Assert as well

packages/jaspr/lib/src/foundation/styles/properties/grid_item.dart
```diff
@@ -77,8 +77,22 @@ class _LinePlacement implements LinePlacement {
   final bool span;
   final String? lineName;
 
+  bool _validLinePlacement() {
+    if (!span && number == null && lineName == null) {
+      throw 'Invalid use of [LinePlacement], cannot have `span == false && number == null && lineName == null`.';
```
schultek: Also can be assert

packages/jaspr/lib/src/foundation/styles/properties/text.dart
```diff
@@ -47,8 +48,19 @@ class _ListFontFamily implements FontFamily {
   final List<FontFamily> families;
   const _ListFontFamily(this.families);
 
+  bool _validFontFamilies() {
+    if (families.isEmpty) {
+      throw '[FontFamily.list] cannot be empty.';
```
schultek: Same here

packages/jaspr/lib/src/foundation/styles/properties/text.dart
```diff
@@ -213,8 +224,18 @@ class _MultiTextDecorationLine implements TextDecorationLine {
 
   const _MultiTextDecorationLine(this.lines);
 
+  bool _validLines() {
+    if (lines.isEmpty) {
+      throw '[TextDecorationLine.multi] cannot be empty.';
```
schultek: Same here

packages/jaspr/lib/src/foundation/styles/properties/text.dart
```diff
@@ -290,13 +311,23 @@ class _TextDecoration implements TextDecoration {
 
   const _TextDecoration({this.line, this.color, this.style, this.thickness});
 
+  bool _attributesNotEntirelyNull() {
+    if (line == null && style == null && color == null && thickness == null) {
+      throw '[TextDecoration] cannot have all attributes null.';
```
schultek: Same here

packages/jaspr/lib/src/foundation/styles/properties/text.dart
```diff
@@ -325,8 +356,25 @@ class _CombineTextShadow implements TextShadow {
 
   final List<TextShadow> shadows;
 
+  bool _shadowsListable() {
+    if (shadows.isEmpty) {
+      throw '[TextShadow.combine] cannot be empty.';
```
schultek: Same here

packages/jaspr/lib/src/foundation/styles/properties/text.dart
```diff
@@ -325,8 +356,25 @@ class _CombineTextShadow implements TextShadow {
...
+    }
+
+    for (final shadow in shadows) {
+      if (shadow is _CombineTextShadow) {
+        throw 'Cannot next [TextShadow.combine] inside [TextShadow.combine].';
```
schultek: This I would allow.

packages/jaspr/lib/src/foundation/styles/properties/transform.dart
```diff
@@ -35,11 +35,34 @@ class _CombineTransform implements Transform {
 
   final List<Transform> transforms;
 
+  bool _transformsListable() {
+    if (transforms.isEmpty) {
+      throw '[Transform.combine] cannot be empty.';
```
schultek, line 40: Assert

packages/jaspr/lib/src/foundation/styles/properties/transform.dart
```diff
@@ -35,11 +35,34 @@ class _CombineTransform implements Transform {
...
+    }
+
+    for (final transform in transforms) {
+      if (transform is _CombineTransform) {
+        throw 'Cannot nest [Transform.combine] inside [Transform.combine].';
```
schultek: Allow

packages/jaspr/lib/src/foundation/styles/properties/transform.dart
```diff
@@ -48,22 +71,32 @@ class _RotateTransform implements Transform {
...
+class _RotateAxisTransform implements _ListableTransform {
   const _RotateAxisTransform({this.x, this.y, this.z});
 
   final Angle? x;
   final Angle? y;
   final Angle? z;
 
+  bool _anglesAvailable() {
+    if (x == null && y == null && z == null) {
+      throw '[Transform.rotateAxis] missing angles (found every angle null).';
```
schultek: Assert

packages/jaspr/lib/src/foundation/styles/properties/transition.dart
```diff
@@ -29,8 +29,25 @@ class _CombineTransition implements Transition {
 
   final List<Transition> transitions;
 
+  bool _transitionsListable() {
+    if (transitions.isEmpty) {
+      throw 'Filter.list cannot be empty.';
```
schultek: Assert

packages/jaspr/lib/src/foundation/styles/properties/box.dart
shubhamlav: Oh, okayyyy, I didn't know that assertions can have an additional argument for message. I'll switch to that. Thanks.

packages/jaspr/lib/src/foundation/styles/properties/box.dart
shubhamlav: Honestly, I don't like the idea of nesting combine with combine. Anyways, so I will remove these. BTW, `Filter.list` also has a check for this. Should I remove it from there as well?

packages/jaspr/lib/src/foundation/styles/properties/transform.dart
shubhamlav: Would it be okay to have asserts at two places? In this case, we still need a check to avoid listing values like none, initial, etc. Having one assertion on the constructor would just spread the logic to a different place and also, it may not be useful at all since `transforms.length > 0` would pass even for invalid values like `[.initial, .revert]`.

shubhamlav:
Huh, interesting. So, this way `Styles` may not always be `const`able. This doesn't sound good. Is there a way around?

schultek:
What can't be const?

shubhamlav:
> What can't be const?
@schultek - This screenshot might be helpful:
image
Basically, I can either access `.length` or I have to give up `const`.

shubhamlav:
BTW, on a different note, do you suggest that I make everything inside the "validate attributes" functions an assert since it conveys a better meaning? i.e.,
```dart
// Current:
bool _validateAttributes() {
  if (someList.isEmpty) {
    throw 'some error';
  }

  for (final e in someList) {
    if (e is Bad) {
      throw 'another error';
    }
  }
}

// More readable (maybe)
bool _validateAttributes() {
  assert(someList.isNotEmpty, 'some error');

  for (final e in someList) {
    assert(e is! Bad, 'another error');
  }
}
```

schultek:
Weird, this works locally for me, but not on dartpad.

shubhamlav:
> Weird, this works locally for me, but not on dartpad.

Have you checked if your style if a final or a const? It works for finals and vars. Only not consts.

schultek:
Ah nvmd, now it complains also locally.

Ok then scratch the asserts on the constructor and lets do everything in the validate method.

> BTW, on a different note, do you suggest that I make everything inside the "validate attributes" functions an assert since it conveys a better meaning? i.e.,

I would do it like this, as you already have:

```dart
bool _validate() {
  if (...) {
    throw 'some error';
  }
  return true;
}

...

assert(_validate());
```

shubhamlav:
@schultek - Any updates?

packages/jaspr/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## Unreleased
+
+- **Breaking**: Changed `FontStyle.obliqueAngle` to accept `Angle` instead of `double`.
+- Allow nesting non-empty `Filter.list` inside each other.
+- Added `initial`, `inherit`, `revert`, `revertLayer` and `unset` to `Transition`.
+- Added `none`, `initial`, `revert`, `revertLayer` and `unset` to `TextShadow`.
```
schultek: Also to `BoxShadow`.

Rest LGTM

shubhamlav: Is this okay or should I move the Breaking change to a new versioning instead of Unreleased Patch? Like
```md
## Unreleased
**Breaking**: Changed `FontStyle.obliqueAngle` ...

## Unreleased Patch
<minor changes go here>
```

schultek (approved):
LGTM

PR #579, opened 2025-10-04
Change double to duration

### LINKED ISSUE #572: feat: support `Duration` for CSS properties instead of `double`
**Description**

It is confusing whether the duration in things like `Transition` is in seconds or milliseconds, and dart has `Duration` so it makes sense to use it.

I propose two things:
- `Duration` to be used `Transition`.
- `DurationExt` (just like `numstr`) to convert that `Duration`'s value into milliseconds.

**Additional Context**

BTW, I can work on this.

ShubhamVG:
Closes #572 

## Description
This adds:
- a getter on `int` to allow shorthand conversion to `Duration`s (in milliseconds)
- an internal extension to convert `Duration` to CSS format milliseconds time. This extension is somewhat needed because there are other places where `Duration` is also needed example `Animation` (coming in my next PR) and possibly CSS variables.
- **Breaking**: `Transition`'s `duration` and `delay` arguments are now `Duration` instead of `double`.

## Type of Change

- ❌ Breaking change

## Ready Checklist

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - Fix tooling daemon crash while analyzing component scopes.
 
+- Added `ms` getter on int to allow shorthand conversion to durations (in milliseconds).
+- **Breaking**: `Transition`'s `duration` and `delay` are now of type `Duration` instead of `double`.
+
 ## 0.21.2
 
 - Added html domain to tooling daemon.
@@ -70,7 +73,7 @@
 
 - Added `allowedPathSuffixes` option to `Jaspr.initializeApp()` to enable handling route paths with extensions other than `html`.
 
-- Added support for generating a **sitemap.xml** in static mode. To enable this, pass `--sitemap-domain=my.domain.com` to `jaspr build`. 
+- Added support for generating a **sitemap.xml** in static mode. To enable this, pass `--sitemap-domain=my.domain.com` to `jaspr build`.
 
   Add sitemap params like `changefreq` and `priority` to your routes by using `Route(settings: RouteSettings(priority: 0.5))` or set them through `ServerApp.requestRouteGeneration()`.
 
@@ -109,7 +112,7 @@
 
 - **BREAKING** Changed `AppBinding`s `Uri get currentUri` to `String get currentUrl`.
 
-- **BREAKING** Changed return type of `renderComponent()` from `Future<String>` to `Future<({int statusCode, String body, Map<String, List<String>> headers})>`. 
+- **BREAKING** Changed return type of `renderComponent()` from `Future<String>` to `Future<({int statusCode, String body, Map<String, List<String>> headers})>`.
 
   The rendered html is accessible through the `body` property. `statusCode` and `headers` can be used to create a response object when part of a custom http handler.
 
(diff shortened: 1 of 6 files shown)


shubhamlav:
I'll need help resolving what to do with dart_quotes_server. I've never used serverpod so I don't have much knowledge about it. Based on what I can understand, serverpod uses SQL somewhere and it is storing duration as `double precision` which now needs to be something else, maybe `INTERVAL`? And even if I change that, what do I have to do to the migrations.sql or definitions.sql?

schultek:
@ShubhamVG  What does this change have to do with serverpod?

shubhamlav:
@schultek Nothing to do with serverpod directly. However, there is one app in the apps/ directory (dart_quotes_server) which needs change since I changed duration from being `double` to `Duration` and for that reason, that app cannot be run anymore, and that is what I need help for. I'm struggling to figure out where to make the changes in order for dart_quotes_server to be compilable/runnable again.

I believe that the SQL stuff in dart_quotes_server is causing this so do I have to make changes to a dataclass or the sql definitions/migration file, I am not sure.

schultek:
I don't think any changes like this are needed there. If the app doesn't run its probably because of something else.

packages/jaspr/lib/src/foundation/styles/properties/unit.dart
```diff
@@ -182,3 +182,13 @@ extension NumberString on double {
     return roundToDouble() == this ? round().toString() : toString();
   }
 }
+
+extension DurationExt on int {
+  /// Shorthand for `Duration(milliseconds: this)`
+  Duration get ms => Duration(milliseconds: this);
+}
+
+extension CssDurationFormat on Duration {
```
schultek: I would remove this extension and directly put the expression in the Transition class, since its only used there.

packages/jaspr/lib/src/foundation/styles/properties/unit.dart
```diff
@@ -182,3 +182,13 @@ extension NumberString on double {
     return roundToDouble() == this ? round().toString() : toString();
   }
 }
+
+extension DurationExt on int {
+  /// Shorthand for `Duration(milliseconds: this)`
+  Duration get ms => Duration(milliseconds: this);
```
schultek, line 188: Can you also add one for seconds? I think thats also quite common.

shubhamlav:
> I don't think any changes like this are needed there. If the app doesn't run its probably because of something else.

@schultek You are right. The error was only a fault on my system. Thanks to Hani from Discord for confirming it. One test is failing because of a generated jaspr_options.dart file but rest seems. There is no serverpod or SQL issue anywhere.

packages/jaspr/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 
 - Fix tooling daemon crash while analyzing component scopes.
 
+- Added `ms` getter on int to allow shorthand conversion to durations (in milliseconds).
+- **Breaking**: `Transition`'s `duration` and `delay` are now of type `Duration` instead of `double`.
+
```
schultek: Move this to the "Unreleased" section, and make it "Unreleased breaking".

schultek: Also check the failing pipeline, thank you :)

shubhamlav:
@schultek Ultimate final ultra pro max commit done (maybe)
Well, at least the tests aren't failing now :saluting_face:
Maybe ready to merge now?

PR #584, opened 2025-10-07
[jaspr_cli] Stricter type linting to avoid unintentional dynamic

parlough:
This updates the analysis to match jaspr_content as I work towards making it consistent across packages.

### DIFF
--- a/packages/jaspr_cli/lib/src/commands/build_command.dart
+++ b/packages/jaspr_cli/lib/src/commands/build_command.dart
@@ -109,7 +109,7 @@ class BuildCommand extends BaseCommand with ProxyHelper, FlutterHelper {
 
     String? entryPoint;
     if (project.requireMode != JasprMode.client) {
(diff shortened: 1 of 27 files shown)


schultek:
Is this still WIP or can be set to ready?

packages/jaspr_cli/lib/src/commands/build_command.dart
```diff
@@ -118,7 +118,7 @@ class BuildCommand extends BaseCommand with ProxyHelper, FlutterHelper {
 
     String? entryPoint;
     if (project.requireMode != JasprMode.client) {
-      entryPoint = await getEntryPoint(argResults!['input']);
```
schultek, line 121: Woow how did I not know about this api.

schultek: I changed this everywhere.

schultek (approved):

PR #598, opened 2025-10-12
Fix compilation errors in `jaspr_riverpod`

simolus3:
## Description

In version `3.0.2` of `package:riverpod`, the type of `Vsync` was changed from `void Function(void Function() task)` to `void Function(Task task)`.

That change broke `jaspr_riverpod`, since `_jasprVsync` does not conform to the changed function type. This applies a fix and raises the dependency on `riverpod` (because the fixed version would not work with riverpod `<3.0.2`).

## Type of Change

🛠️ Bug fix

## Ready Checklist

- [X] I added myself to the AUTHORS file (optional, if you want to).

### DIFF
--- a/packages/jaspr_riverpod/lib/src/core/provider_scope.dart
+++ b/packages/jaspr_riverpod/lib/src/core/provider_scope.dart
@@ -322,7 +322,15 @@ class _UncontrolledProviderScopeElement extends InheritedElement {
   }
 
   void _jasprVsync(Task task) {
-    assert(_task == null, 'Only one task can be scheduled at a time');
+    assert(
+      _task == null ||
+          // Checks for race conditions where a task has been completed in a different scope.
+          // If so, it then becomes possible for another task to be scheduled
+          // before this scoped had the opportunity to run its task.
+          // TODO find a way to test this.
+          _task!.completed,
+      'Only one task can be scheduled at a time',
+    );
     _task = task;
 
     Future.microtask(() async {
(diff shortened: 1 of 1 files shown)


packages/jaspr_riverpod/lib/src/core/provider_scope.dart
```diff
@@ -321,7 +321,7 @@ class _UncontrolledProviderScopeElement extends InheritedElement {
     setDependencies(dependent, getDependencies(dependent) ?? ProviderDependencies(dependent, this));
   }
 
-  void _jasprVsync(void Function() task) {
+  void _jasprVsync(Task task) {
     assert(_task == null, 'Only one task can be scheduled at a time');
```
schultek: flutter_riverpod changed this assert to:

```dart
assert(
      _task == null
          // Checks for race conditions where a task has been completed in a different scope.
          // If so, it then becomes possible for another task to be scheduled
          // before this scoped had the opportunity to run its task.
          // TODO find a way to test this.
          ||
          _task!.completed,
      'Only one task can be scheduled at a time',
    );
```

and we should do the same

packages/jaspr_riverpod/lib/src/core/provider_scope.dart
schultek, line 1: Also we should change the callsite of task (Line 337) to reflect the changes of flutter_riverpod:

```dart
final task = _task;
_task = null;
task?.call();
```

simolus3:
Thanks for completing this 👍

PR #603, opened 2025-10-14
fix whitespace rendering and improve comments

### LINKED ISSUE #602: [content] Markdown inline spans nested within blocks render with extra spaces
When at the top level, this Markdown renders as expected:

```markdown
`abstract` with `sealed`. A [sealed](#sealed) class is
implicitly [abstract](#abstract).

`interface`, `final` or `sealed` with `mixin`. These access modifiers
prevent [mixing in][mixin].
```

However, when rendered within a list element, like:

```markdown
* `abstract` with `sealed`. A [sealed](#sealed) class is
  implicitly [abstract](#abstract).
* `interface`, `final` or `sealed` with `mixin`. These access modifiers
  prevent [mixing in][mixin].
```

There is an extra space added after each inline span and before the periods:

Screenshot of the rendered output

schultek:
## Description

Fixes #602 

## Type of Change

- 🛠️ Bug fix

### DIFF
--- a/packages/jaspr/lib/src/server/markup_render_object.dart
+++ b/packages/jaspr/lib/src/server/markup_render_object.dart
@@ -71,7 +71,7 @@ abstract class MarkupRenderObject extends RenderObject {
     var trailingWhitespace = false;
 
     if (children.isNotEmpty) {
-      final childOutput = <(String, bool, bool, bool)>[];
+      final childOutput = <({String html, bool leading, bool trailing, bool hasNewline})>[];
       var childOutputLength = 0;
       var childOutputLinebreak = false;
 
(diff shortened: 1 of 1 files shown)


packages/jaspr/lib/src/server/markup_render_object.dart
```diff
@@ -57,13 +64,14 @@ abstract class MarkupRenderObject extends RenderObject {
...
     final output = StringBuffer();
+    // Don't increase the indent for fragments.
     final childIndent = this is MarkupRenderFragment ? indent : '$indent  ';
 
     var leadingWhitespace = false;
     var trailingWhitespace = false;
 
     if (children.isNotEmpty) {
-      final childOutput = <(String, bool, bool)>[];
+      final childOutput = <(String, bool, bool, bool)>[];
```
parlough: Consider changing these records to use named fields, as the code later on with the indexed accesses is a bit hard to understand without referring back.

schultek: Agreed

Issue #604, opened 2025-10-15
[content] Data based override to disable template processing for a `Page`


parlough:
Can there be a way for an individual page (through its frontmatter/data) to skip rendering with the configured template engine?

I know I can create a separate `PageConfig` but it doesn't always makes sense when everything else will be shared.

schultek:
Since the `TemplateEngine` class gets the full page, it can decide when to render or not. So I would just recommend subclassing `TemplateEngine` and checking for some page data value yourself. This doesn't need a change in jaspr_content imo.

Edit: Reason is I don't want to add too many "magic" data properties in jaspr_content itself.

PR #605, opened 2025-10-16
[content] Enable a page to skip template rendering

parlough:
Resolves https://github.com/schultek/jaspr/issues/604


schultek:
See comment on issue

PR #611, opened 2025-10-22
feat: html class components

### LINKED ISSUE #590: [feature] Explore making html component functions classes with const constructors
While they don't follow typical naming patterns for Dart classes/types, the distinction serves a purpose that is potentially worth the one time exception.

schultek:
## Description

- Refactor all html components to be class components instead of functions.
- Move all DOM APIs (html components, styles, etc.) to `package:jaspr/dom.dart` library.
- Deprecate `text()`, `fragment()` and `raw()` functions.
- Remove `package:jaspr/ui.dart` library.
- Update all apps and examples to Dart 3.10 and migrate to dot shorthands.

Resolves #590 

## Type of Change

- ❌ Breaking change

## Ready Checklist


packages/jaspr/lib/dom.dart
```diff
@@ -0,0 +1,7 @@
+export 'jaspr.dart' hide runApp, Document;
```
parlough: I'd personally recommend dropping this export. It will cause inconsistencies in how people access members from `jaspr.dart`.

It'll also result in `dom.dart` showing up as a duplicate option when adding imports for items from `jaspr.dart`, requiring developers to have to keep making extra choices.

packages/jaspr/lib/ui.dart
parlough, line 1: Thanks for removing this. The utilities were nice, but dropping them simplifies Jaspr and developers can replicate the functionality relatively easily.

packages/jaspr/lib/src/dom/validator.dart
parlough, line 1: Considering this restructuring, what do you think about making this an internal class and not exporting it in the public API? It seems like it might need to evolve over time and is more catered towards Jaspr's internal usage. 

If you think it's valuable outside of Jaspr, perhaps we can make a more general purpose package later on.

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -2,8 +2,8 @@ part of 'framework.dart';
 
 /// Represents a html element in the DOM
 ///
-/// Must have a [tag] and any number of attributes.
-/// Can have a single [child] component or any amount of [children].
+/// Must have a [tag] and any number of [attributes].
+/// Accepts a list of [children].
```
parlough, line 6: Nice catch on this old mention.

But it makes me wonder, should this and `Component.element` accept a single child so they're consistent with self-authored components? Especially now that `.fragment` is easier with this change?

The html-classes like `div` could of course still accept a list to be consistent with the html experience they provide.

packages/jaspr/CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 ## Unreleased breaking
 
+- **Breaking** Moved all html components, style classes and dom utilities including `div()` et al., `Styles`, `css`, `Color` et al., `events()`, `RawText`, `DomValidator` and `ViewTransitionMixin` to separate `package:jaspr/dom.dart` library.
+
+  This reduces the "pollution" of the global namespace when importing `package:jaspr/jaspr.dart` and allows for more fine-grained control of imported APIs.
```
parlough, line 5: Woo!! Thank you 🥳

packages/jaspr/tool/generate_html.dart
```diff
@@ -18,189 +23,333 @@ const htmlSpec = ${const JsonEncoder.withIndent('  ').convert(specJson)};
...
-            final name = type['name'];
-            content.write(name);
-          } else if (type == 'content') {
-            content.write('String');
-            contentParam = name;
-          } else {
-            throw ArgumentError('Attribute type is unknown ($type) for attribute $key.$tag.$attr');
-          }
+        l.body.add(
+          Class((c) {
```
parlough, line 50: Consider adding the `final` [modifier](https://pub.dev/documentation/code_builder/latest/code_builder/Class/modifier.html) to each of these generated classes. They definitely shouldn't be extended or implemented by Jaspr developers :D 

That should also prevent it from ever showing up as a confusing auto-completion suggestion when implementing or extending a class.

packages/jaspr/lib/src/framework/components.dart
schultek: That would mean all html components would need to build an additional fragment, almost doubling the total number of render objects in an app and causing a potentially huge drop in performance. Also I still want this api to represent standard html elements, which naturally can have multiple children.

packages/jaspr/tool/generate_html.dart
schultek: Good idea

packages/jaspr/lib/dom.dart
schultek: Sounds fair.

packages/jaspr/lib/src/dom/validator.dart
schultek: I removed it from the library exports.

PR #614, opened 2025-10-23
Remove dependency on jaspr_repo from published packages

parlough:
This should hopefully allow analysis on pub.dev to succeed.


schultek:
@parlough This is working for me locally and the pipeline also succeeds. I'm happy to merge when you set it to ready.

PR #627, opened 2025-11-03
[content] Enable customizing header of secondary markdown output

parlough:


parlough:
> How about making this configurable through the constructor. Either as a bool or even 'String Function(Page) createHeader' parameter.

Good idea, especially since a `title` or `description` isn't necessarily required for a `Page`.

I implemented the function parameter, let me know what you think. Thank you!

PR #641, opened 2025-11-10
[jaspr_riverpod] Update to use stricter analysis

parlough:
Updates the analysis config for `package:jaspr_riverpod` to match the other packages on the road to using a shared, stricter config across all the packages.


schultek (approved):
Thanks for pushing this for all the packages

PR #652, opened 2025-11-14
Update core package to stricter type analysis config

parlough:
Updates the analysis config for `packages/jaspr` to use stricter type analysis to match the other packages as we work towards unifying the analysis for the repository.

### DIFF
--- a/packages/jaspr/test/server/client_test.dart
+++ b/packages/jaspr/test/server/client_test.dart
@@ -34,7 +34,7 @@ class _App3State extends State<App3> with SyncStateMixin<App3, Map<String, Objec
   int count = 3;
 
   @override
-  Map<String, int> getState() {
+  Map<String, Object?> getState() {
     return {'count': count};
   }
 
@@ -66,7 +66,7 @@ class _App4State extends State<App4> with SyncStateMixin<App4, Map<String, Objec
   }
 
   @override
-  Map<String, String> getState() {
+  Map<String, Object?> getState() {
     return {'label': label};
   }
 
--- a/packages/jaspr/tool/generate_html.dart
+++ b/packages/jaspr/tool/generate_html.dart
@@ -183,10 +183,11 @@ const htmlSpec = ${const JsonEncoder.withIndent('  ').convert(specJson)};
       content.write('    events: ');
 
       if (events.isNotEmpty) {
+        final needsTypeArgs = events.length == 1 && events.first == 'onClick';
         content.write(
           '{\n'
           '      ...?events,\n'
-          '      ..._events(${events.map((e) => '$e: $e').join(', ')}),\n'
+          '      ..._events${needsTypeArgs ? '<void, void>' : ''}(${events.map((e) => '$e: $e').join(', ')}),\n'
           '    },\n',
         );
       } else {


packages/jaspr/lib/src/components/html/forms.dart
```diff
@@ -33,7 +33,7 @@ Component button(
     },
     events: {
       ...?events,
-      ..._events(onClick: onClick),
```
schultek, line 36: I think the `events` method should be marked with `@optionalTypeArgs` instead, providing the types here and at the other call sites is not nice.

packages/jaspr/lib/src/foundation/sync.dart
```diff
@@ -28,7 +28,7 @@ mixin SyncStateMixin<T extends StatefulComponent, U> on State<T> {
     initSyncState(this);
   }
 
-  static initSyncState(SyncStateMixin mixin) => s.initSyncState(mixin);
```
schultek: I'd rather have this non-generic and accept `SyncStateMixin<StatefulComponent, Object?>`

packages/jaspr/lib/src/foundation/sync/sync_vm.dart
```diff
@@ -2,16 +2,16 @@ import 'dart:convert';
 
 import '../../../server.dart';
 
-void initSyncState(SyncStateMixin element) {
```
schultek, line 5: Same here, keep non-generic

packages/jaspr/lib/src/foundation/sync/sync_vm.dart
```diff
@@ -2,16 +2,16 @@ import 'dart:convert';
...
+void initSyncState<T extends StatefulComponent, U>(SyncStateMixin<T, U> element) {
   if (element.context.binding case ServerAppBinding b) {
     b.addRenderAdapter(SyncAdapter(element, element.context as Element));
   }
 }
 
-class SyncAdapter extends ElementBoundaryAdapter {
```
schultek, line 11: Same here

packages/jaspr/lib/src/foundation/sync/sync_web.dart
```diff
@@ -5,7 +5,7 @@ import '../../foundation/type_checks.dart';
 
 final _syncRegex = RegExp('^${DomValidator.syncMarkerPrefixRegex}(.*)\$');
 
-void initSyncState(SyncStateMixin sync) {
```
schultek, line 8: same here

packages/jaspr/test/server/client_test.dart
```diff
@@ -30,17 +30,17 @@ class App3 extends StatefulComponent {
   State<App3> createState() => _App3State();
 }
 
-class _App3State extends State<App3> with SyncStateMixin {
```
schultek, line 33: `Map<String, Object?>`

packages/jaspr/test/server/client_test.dart
```diff
@@ -56,23 +56,23 @@ class App4 extends StatefulComponent {
   State<App4> createState() => _App4State();
 }
 
-class _App4State extends State<App4> with SyncStateMixin, PreloadStateMixin {
```
schultek, line 59: `Map<String, Object?>`

packages/jaspr/lib/src/components/html/forms.dart
parlough: Good idea. I've added the annotation. However, this `_events` used for within the library doesn't propagate the annotation. I've kept the `<void, void>` for these two use cases for now but I'll file an issue upstream so that the we can perhaps add the annotation to the variable later on.

parlough:
Thanks for the feedback @schultek! I updated the PR according to your comments, PTAL.

One exception is https://github.com/schultek/jaspr/pull/652#discussion_r2535164823. While I've added `@optionalTypeArgs` to `events`, some more changes are required to drop the `<void, void>` on our two usages of it, so I've kept them for now.

parlough:
Updated with the latest `main` to resolve conflicts. Ready for another look when you get a chance. Thanks :D

PR #654, opened 2025-11-16
fixes custom body responses having a html extension

Lexedia:
## Description

This is a followup on #597 that sets an approximate extension when using a custom body response.

## Type of Change

- 🛠️ Bug fix

## Ready Checklist


schultek:
Can you elaborate why this is needed?

You can already define Routes with an extension, e.g. `Route(path: 'some/file.json', ...)`.

Lexedia:
Well, it was mostly for routes without an extension set, although forcing one can also work indeed.

Tbh, it was mostly my fried brain at 2 o'clock that decided to make this PR

PR #661, opened 2025-11-24
[jaspr_pad] Update `jaspr_riverpod` dependency constraint

parlough:


apps/jaspr_pad/samples/riverpod/main.dart
```diff
@@ -23,10 +22,17 @@ class App extends StatelessComponent {
...
     ]);
   }
 }
+
+class _CounterNotifier extends Notifier<int> {
+  @override
+  int build() => 0;
+
+  void increment() => state = state + 1;
+}
```
schultek: I would move this ontop of `App` below the provider definition.

schultek (approved):

PR #696, opened 2025-12-08
[jaspr] Add dom components for description list elements

parlough:


parlough:
@schultek Would it still be possible for these additions to make it into the next stable release?

PR #717, opened 2025-12-29
Fix updating event handlers

abitofevrything:
## Description

#652 [changed the way that event handlers are registered on DOM elements in the client](https://github.com/schultek/jaspr/pull/652/changes#diff-00f6d28448c0d0801bb9d857258dbbf9a59faa58bbf3cf311dcd1b4fb9de2a1aL568). Unfortunately this small change made it so that the `fn` identifier in the callback no longer points to the field on the `EventBinding`, but rather the parameter to its constructor. This then means that later updating the callback by setting `EventBinding.fn` to the new callback doesn't actually update the callback that is invoked upon an event firing. This PR reverts that part of the change.

This causes all sorts of issues, notably that event handlers that refer to variables declared in the surrounding `build` method aren't called with up to date values for those variables when a rebuild happens.

I'd therefore like to request this PR to be a candidate for a patch release.

## Type of Change

- 🛠️ Bug fix

## Ready Checklist

### DIFF
--- a/packages/jaspr/lib/src/client/dom_render_object.dart
+++ b/packages/jaspr/lib/src/client/dom_render_object.dart
@@ -597,10 +597,10 @@ class EventBinding {
 
   EventBinding(web.Element element, this.type, this.fn) {
     _subscription = web.EventStreamProvider<web.Event>(type).forElement(element).listen((event) {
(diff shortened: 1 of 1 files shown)


schultek (approved):
Thanks for catching this and sending a PR! LGTM

PR #718, opened 2025-12-31
Support riverpod versions 3.1.0+

simolus3:
## Description

Since https://github.com/rrousselGit/riverpod/commit/b945ca09345d3af11f5547dbb08073cbd8d7dcea, `$FunctionalProvider` no longer has a constant constructor. This causes a compilation error in `jaspr_riverpod` for apps depending on `riverpod` version `3.1.0`. This fixes that by also removing the `const` here.

## Type of Change

🛠️ Bug fix


schultek (approved):
Thanks for keeping this up to date. LGTM

PR #727, opened 2026-01-11
fix: support RawText components in testComponents

### LINKED ISSUE #726: fix: testComponents fails with RawText (SVG) components - TestRenderElement cast error
# Bug: `testComponents` fails with RawText (SVG) components - TestRenderElement cast error

## Description

When using `testComponents` from `jaspr_test` to test components that contain `RawText` (typically used for inline SVG icons), the test fails with a type cast error. This prevents testing any component that includes SVG icons or other raw HTML/XML content.

## Steps to Reproduce

1. Create a component that uses `RawText` for SVG icons:

```dart
// svg_icons.dart
import 'package:jaspr/jaspr.dart';

class SvgIcons {
  static Component menu() => RawText('''
    <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor">
      <line x1="3" y1="12" x2="21" y2="12"/>
      <line x1="3" y1="6" x2="21" y2="6"/>
      <line x1="3" y1="18" x2="21" y2="18"/>
    </svg>
  ''');
}
```

2. Use this in a component:

```dart
// navigation.dart
import 'package:jaspr/jaspr.dart';
import 'svg_icons.dart';

class Navigation extends StatelessComponent {
  @override
  Component build(BuildContext context) {
    return nav([
      button([SvgIcons.menu()]),
    ]);
  }
}
```

3. Write a test:

```dart
import 'package:jaspr_test/jaspr_test.dart';
import 'navigation.dart';

void main() {
  testComponents('renders navigation', (tester) async {
    tester.pumpComponent(Navigation());

    expect(find.tag('nav'), findsOneComponent);
  });
}
```

4. Run `dart test`

## Expected Behavior

The test should render the component tree (including RawText elements) in the simulated environment and allow assertions.

## Actual Behavior

Test fails immediately with the following error:

```
type 'TestRenderElement' is not a subtype of type 'MarkupRenderObject' in type cast
package:jaspr/src/dom/raw_text/raw_text_vm.dart 28:60 \_RawTextElement.createRenderObject
package:jaspr/src/framework/render_object.dart 47:23 RenderObjectElement.didMount
...
```

## Environment

- jaspr: 0.22.1
- jaspr_test: 0.22.1
- Dart SDK: 3.10.0
- OS: macOS

## Analysis

The issue is in `raw_text_vm.dart` at line 28:

```dart
class _RawTextElement extends LeafRenderObjectElement {
  _RawTextElement(super.component);

  @override
  RenderObject createRenderObject() {
    final parent = parentRenderObjectElement!.renderObject as MarkupRenderObject;  // <-- Cast fails here
    return parent.createChildRenderText((component as RawText).text, true);
  }
  // ...
}
```

When `testComponents` is used, `parentRenderObjectElement!.renderObject` returns a `TestRenderObject`, but the code casts it to `MarkupRenderObject`. Since `TestRenderObject` doesn't extend `MarkupRenderObject`, the cast fails.

**Root cause**: The VM implementation of `RawText` assumes it will always run in a server context with `MarkupRenderObject`, but `testComponents` uses `TestRenderObject` hierarchy.

## Suggested Fix

The fix should make `_RawTextElement.createRenderObject()` work with any `RenderObject` that supports `createChildRenderText`:

```dart
// Option 1: Use the base RenderObject interface
class _RawTextElement extends LeafRenderObjectElement {
  _RawTextElement(super.component);

  @override
  RenderObject createRenderObject() {
    final parent = parentRenderObjectElement!.renderObject;
    return parent.createChildRenderText((component as RawText).text, true);
  }

  @override
  void updateRenderObject(covariant RenderObject renderObject) {
    if (renderObject is RenderText) {
      renderObject.update((component as RawText).text, true);
    }
  }
}
```

However, the base `RenderObject.createChildRenderText()` only accepts one parameter (`String text`), while `MarkupRenderObject.createChildRenderText()` accepts two (`String text, [bool rawHtml = false]`).

**The proper fix** would be one of:

1. **Add `rawHtml` parameter to the base `RenderText` interface** in `framework.dart`:

```dart
abstract class RenderText implements RenderObject {
  void update(String text, [bool rawHtml = false]);
}
```

2. **Add a `TestRenderRawText` class** in `jaspr_test` that handles raw HTML rendering in tests.

3. **Make `_RawTextElement` check the runtime type** and handle both cases:

```dart
@override
RenderObject createRenderObject() {
  final parent = parentRenderObjectElement!.renderObject;
  final text = (component as RawText).text;

  if (parent is MarkupRenderObject) {
    return parent.createChildRenderText(text, true);
  }
  // Fallback for test environment
  return parent.createChildRenderText(text);
}
```

## Workaround

Currently, the only workaround is to avoid testing components that contain `RawText`. We've resorted to testing only provider/state logic with standard unit tests instead of component tests.

BiAksoy:
### Description

This PR fixes the `testComponents` test failure when testing components that contain `RawText` (typically used for inline SVG icons).

When using `testComponents` from `jaspr_test`, components with `RawText` fail with:

```
type 'TestRenderElement' is not a subtype of type 'MarkupRenderObject' in type cast
```

**Root cause:** The `_RawTextElement.createRenderObject()` method in `raw_text_vm.dart` was casting the parent render object to `MarkupRenderObject`, but in the test environment, `TestRenderObject` is used instead, causing the cast to fail.

**Solution:**

- Add optional rawHtml parameter to RenderObject.createChildRenderText()
- Add optional rawHtml parameter to RenderText.update()
- Remove MarkupRenderObject cast in raw_text_vm.dart
- Update TestRenderObject and TestRenderText for rawHtml support

Fixes #726 

### Type of Change

- [x] 🛠️ Bug fix

### Ready Checklist

- [x] All tests pass (`melos run test`)
- [x] No breaking changes

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,8 +1,9 @@
 ## Unreleased patch
 
 - Fixed `testComponents` failing with `RawText` components due to `MarkupRenderObject` cast error.
-  - Added optional `rawHtml` parameter to `RenderObject.createChildRenderText()` and `RenderText.update()` interfaces.
-  - Updated `TestRenderObject` and `TestRenderText` to support raw HTML text rendering in tests.
+  - Added `RawableRenderObject` and `RawableRenderText` interfaces for raw HTML text rendering support.
+  - `MarkupRenderObject` and `TestRenderObject` now implement `RawableRenderObject`.
+  - `MarkupRenderText` and `TestRenderText` now implement `RawableRenderText`.
 
 ## 0.22.1
 
--- a/packages/jaspr/lib/src/client/dom_render_object.dart
+++ b/packages/jaspr/lib/src/client/dom_render_object.dart
@@ -27,7 +27,7 @@ abstract class DomRenderObject implements RenderObject {
   }
 
   @override
-  RenderText createChildRenderText(String text, [bool rawHtml = false]) {
+  RenderText createChildRenderText(String text) {
     return DomRenderText(text, this);
   }
 
@@ -250,7 +250,7 @@ class DomRenderText extends DomRenderObject implements RenderText {
   }
 
   @override
-  void update(String text, [bool rawHtml = false]) {
+  void update(String text) {
     if (node.textContent != text) {
       node.textContent = text;
       if (kVerboseMode) {
--- a/packages/jaspr/lib/src/dom/raw_text/raw_text_vm.dart
+++ b/packages/jaspr/lib/src/dom/raw_text/raw_text_vm.dart
@@ -25,12 +25,12 @@ class _RawTextElement extends LeafRenderObjectElement {
 
   @override
   RenderObject createRenderObject() {
-    final parent = parentRenderObjectElement!.renderObject;
+    final parent = parentRenderObjectElement!.renderObject as RawableRenderObject;
     return parent.createChildRenderText((component as RawText).text, true);
   }
 
   @override
   void updateRenderObject(covariant RenderObject renderObject) {
-    (renderObject as RenderText).update((component as RawText).text, true);
+    (renderObject as RawableRenderText).update((component as RawText).text, true);
   }
 }
--- a/packages/jaspr/lib/src/framework/render_object.dart
+++ b/packages/jaspr/lib/src/framework/render_object.dart
@@ -5,7 +5,7 @@ abstract class RenderObject {
   web.Node? get node;
 
   RenderElement createChildRenderElement(String tag);
-  RenderText createChildRenderText(String text, [bool rawHtml = false]);
+  RenderText createChildRenderText(String text);
   RenderFragment createChildRenderFragment();
 
   void attach(covariant RenderObject child, {covariant RenderObject? after});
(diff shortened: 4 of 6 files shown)


schultek (commented):
Thanks for sending this PR, it's a great fix but I would take a slightly different approach. 

Mainly I want to avoid `DomRenderObject` having the `rawHtml` parameter because it's not used. We should make a separate interface class (e.g. RawableRenderObject) that both `MarkupRenderObject` and `TestRenderObject` implement and that has this parameter. Then in `RawText` we can cast to this interface.

BiAksoy:
Thanks for the feedback! I've updated the PR with the suggested approach:

Created two new interface classes:
- `RawableRenderObject` - for render objects that can create raw HTML text children
- `RawableRenderText` - for text render objects that support raw HTML updates

Now:
- `MarkupRenderObject` implements `RawableRenderObject`
- `TestRenderObject` implements `RawableRenderObject`  
- `MarkupRenderText` implements `RawableRenderText`
- `TestRenderText` implements `RawableRenderText`
- `DomRenderObject` stays unchanged (no unnecessary rawHtml parameter)
- `raw_text_vm.dart` casts to these interfaces

This way, `DomRenderObject` stays clean and RawText works.

All tests pass. Let me know if any further changes are needed!

PR #735, opened 2026-01-27
Add cross-compilation to build command

### LINKED ISSUE #714: Cross compilation
From Discord: https://discord.com/channels/993167615587520602/1454537012634849381

On server mode, `jaspr build` is generating an `app` executable file with the following info:

```
app: Mach-O 64-bit executable arm64
```

But I need it to be targeted for Linux X64. In Dart I do this in a simple way, as cross compilation is available.

What about jaspr?

zeykafx:
## Description

This PR adds cross-compilation of the server app. 
To do this, two flags are added to the `jaspr build` command: `--target-os` and `--target-arch`.

A Linux x86-64 binary can be compiled from MacOS (arm64) (or Windows) using the following command

```shell
jaspr build --target exe \
  --target-os linux \
  --target-arch x64
```
Binary file info:
```
$ file ./build/jaspr/app
./build/jaspr/app: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=13e2183ee72b58941d9005e46196852ba0d96a1f, stripped
```

I successfully tested the cross-compilation from MacOS to Linux x86-64 and Linux arm64 on virtual machines.

Note: Dart currently only supports cross-compilation to Linux, as documented [here](https://dart.dev/tools/dart-compile#cross-compilation-exe)

Fixes #714.

I also added a small section about cross-compilation in the documentation.
The change was also added to the changelog.

## Type of Change

- 📝 Documentation

## Ready Checklist

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased patch
 
+- Added `--web-port` and `--proxy-port` options to `jaspr serve` for configuring webdev and proxy server ports.
+  - Allows running multiple Jaspr projects simultaneously with different ports.
+
 - Fixed `testComponents` failing with `RawText` components due to `MarkupRenderObject` cast error.
   - Added `RawableRenderObject` and `RawableRenderText` interfaces for raw HTML text rendering support.
   - `MarkupRenderObject` and `TestRenderObject` now implement `RawableRenderObject`.
@@ -8,6 +11,9 @@
 - Fixed encoding of `lastmod` property in generated sitemap to be a valid W3C date string.
 
 - Added `--target-os` and `--target-arch` options to the build command to cross-compile server binaries.
+- Added `--dart-define-from-file` option to the serve and build commands.
+
+- Added an example to show on pub.dev.
 
 ## 0.22.1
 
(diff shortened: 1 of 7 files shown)


packages/jaspr_cli/lib/src/commands/build_command.dart
```diff
@@ -158,6 +168,9 @@ class BuildCommand extends BaseCommand with ProxyHelper, FlutterHelper {
     if (project.requireMode == JasprMode.server) {
       logger.write('Building server app...', progress: ProgressState.running);
 
+      final String? targetOS = argResults!.option('target-os');
+      final String? targetArch = argResults!.option('target-arch');
+
       final compileTarget = argResults!.option('target')!;
       final extension = switch (compileTarget) {
         'exe' when Platform.isWindows => '.exe',
```
schultek: I guess when cross compiling from windows to linux, we shouldn't add the `.exe` suffix.

schultek:
You can ignore the test warning, I think manual testing is enough here.

packages/jaspr_cli/lib/src/commands/build_command.dart
schultek: Done

zeykafx:
Ah great, thanks!! Sorry for not fixing it in time!

schultek:
No worries, just didn't want it to miss the release

PR #737, opened 2026-01-30
Hosted pub

xclud:
## Description

Respect `PUB_HOSTED_URL` for hosted pub mirrors since pub.dev is banned in some countries like china.

More docs: https://docs.flutter.dev/community/china

https://github.com/VeryGoodOpenSource/pub_updater/blob/fdec17aea4c65948488f6e9a0f0bfa804dbae342/lib/src/pub_updater.dart#L16

## Type of Change

✨ New feature or improvement

## Ready Checklist

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 - Fixed encoding of `lastmod` property in generated sitemap to be a valid W3C date string.
 
+- Added `--dart-define-from-file` option to `serve` and `build` commands.
+
 - Added an example to show on pub.dev.
 
 - Respect `PUB_HOSTED_URL` environment variable for custom pub mirrors.
--- a/packages/jaspr_cli/lib/src/helpers/dart_define_helpers.dart
+++ b/packages/jaspr_cli/lib/src/helpers/dart_define_helpers.dart
@@ -1,4 +1,28 @@
+import 'dart:convert';
+import 'dart:io';
+
 import '../commands/base_command.dart';
+
+/// Regex patterns for parsing .env files (matches Flutter's implementation).
+abstract class _DotEnvRegex {
+  // Dot env multi-line block value regex (to reject)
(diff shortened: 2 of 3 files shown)


packages/jaspr_cli/lib/src/commands/create_command.dart
```diff
@@ -125,7 +126,7 @@ class CreateCommand extends BaseCommand {
       if (useMode.useServer) useBackend ?? 'base',
     };
 
-    final updater = PubUpdater();
+    final updater = PubUpdater(null, getPubDevBaseUrl());
```
schultek: To reduce duplication, can you add this as a final field to `BaseCommand` and use that in every command.

packages/jaspr_cli/lib/src/utils.dart
```diff
@@ -0,0 +1,16 @@
+import 'dart:io';
+
+const _PUB_HOSTED_URL_ENV_NAME = 'PUB_HOSTED_URL';
+
+/// Gets the base url for pub mirror with respect to the value of `PUB_HOSTED_URL` enviroment variable.
+///
+/// Falls back to https://pub.dev if the enviroment variable is not found.
+String getPubDevBaseUrl() {
+  for (final kv in Platform.environment.entries) {
+    if (kv.key.toUpperCase() == _PUB_HOSTED_URL_ENV_NAME) {
```
schultek: So it is important to also allow lowercase? Why not just check `Platform.environment['PUB_HOSTED_URL']`?

xclud: Based on my experience users define the environment variables diffrently on diffrent OSes e.g. Linux or Windows.

That's why i am not using `[]` indexer to check for both lower and upper cases.

xclud: I've chaged as you proposed. Also fixed a lint error.

packages/jaspr_cli/lib/src/commands/create_command.dart
xclud: Done as you proposed.

PR #745, opened 2026-02-11
Add asset manager

### LINKED ISSUE #446: feat: [content] Content assets are automatically added (aka "page bundle")
I think any tool that allows SSG style of building like jaspr_content should have means to have additional assets directly next to markdown files, since these assets (mostly images) are also part of the content, and should not leave the folder therefore. I think those are often called "page bundles", i.e. in Hugo. I imagine it as changing SSG should not affect content directory much, except maybe for custom components which are handled a bit differently in each.

---

I propose to create a mechanism that would automatically put all stuff in content (probably without md files but not sure since maybe they can also be used as assets to be downloaded?) should be processed, maybe create unique name based on its path (/content/posts/post1_image.jpg to i.e. content--posts--post1_image.jpg or copies to special directory that can be specified ie. /public/posts/post1_image.jpg, does not really matter) and keep this list for further processing, lets call it maybe asset list. Maybe when you would use image from static assets (from web/), there should be no more processing.

Then there will be page extension that would basically scan for used assets, i.e. image, and update it's path to the one constructed using that asset list, therefore it would link it to correct path and therefore image path (or path in links in general) would be transformed into new endoced path from asset list.

There should be also a way to hook into that and apply modifications to the specific link (image) used, so I can set how the concrete image is transformed (i.e. reduced size or dimensions). That would have to create another entry in the asset list since it basically creates new image.

Each entry in asset list should be using page extension determined whether it is used -- and IDEALLY (maybe possible to done in future) if some transformation that generates new asset is used, the original should not be counted as used if not used without transformation in other place.

This way we can only produce to build those assets that are only really used.

In general, there also might be a function that would add any provided asset from within whatever part of Jaspr.

---

For reference, in Hugo there is https://gohugo.io/content-management/page-bundles/. But in Hugo and other SSG like Jekyll etc. the folder structure is also url, which is not true with Jaspr necessary.

Maybe better example would be Astro which from my minimal knowledge supports assets in /public/ directly (static files) and also from source, which can be processed/transformed. https://docs.astro.build/en/guides/images/#using-local-images-in-mdx-or-markdown

---

There should be some evaluation whether this logic is correct and some more deep technical ideas, but currently I don't have deep Jaspr knowledge to write here more.

schultek:
## Description

Adds `AssetManager` to `jaspr_content` for handling local assets.

Resolves #446 

## Type of Change

## Ready Checklist

### DIFF
--- a/packages/jaspr_content/CHANGELOG.md
+++ b/packages/jaspr_content/CHANGELOG.md
@@ -7,8 +7,7 @@
 
 - Added `filterExtensions` parameter to `FilesystemLoader`, which allows filtering for specific file extensions. This can be used when co-locating assets and data files with the content to prevent them from being treated as pages.
 
-- Fixed `FilesystemDataLoader` to ignore files it cannot parse. This allows for co-locating data files with the content.
-
+- **Breaking** `FilesystemDataLoader` now ignores files it cannot parse (non json or yaml files), instead of returning a `String`.
 - **Breaking**: `RouteLoaderBase` now accepts a generic type parameter for specifying the respective `PageSource` subclass.
 - **Breaking**: `Page.path` is now always in posix format (using forward slashes), independent of the operating system.
 
--- a/docs/content/concepts/data_loading.mdx
+++ b/docs/content/concepts/data_loading.mdx
@@ -40,7 +40,12 @@ The `FilesystemDataLoader` loads data from files in a directory of your project.
 
 The files are parsed based on their extension:
 - **.yaml** and **.json** files are loaded and parsed into a `Map<String, dynamic>`
-- all other files are loaded as a raw `String`
+- all other files are **ignored**.
+
+<Info>
+
+You can also set the data directory to the same directory as your content files, which allows co-locating data with your content. In that case, you should use the `filterExtensions` parameter on [`FilesystemLoader`](/content/concepts/route_loading#filesystemloader) to prevent the data files from being treated as pages.
+</Info>
 
 #### Directory Structure to Data Access
 
--- a/docs/content/concepts/route_loading.mdx
+++ b/docs/content/concepts/route_loading.mdx
@@ -33,6 +33,7 @@ The `FilesystemLoader` loads pages from a specified directory on your local file
 - **Features**:
   - Supports hot-reloading in debug mode: Changes to files trigger a rebuild of the affected pages.
   - Can keep file suffixes for specific paths using the `keepSuffixPattern` argument.
+  - Can ignore unknown file extensions using the `filterExtensions` argument, which allows for co-locating data files or assets within the content directory.
 
 **Example:**
 
--- a/packages/jaspr_content/test/data_loaders/filesystem_data_loader_test.dart
+++ b/packages/jaspr_content/test/data_loaders/filesystem_data_loader_test.dart
@@ -84,6 +84,36 @@ void main() {
       );
     });
 
+    test('ignores unsupported file types', () async {
+      // Arrange
+      final dataDir = fileSystem.directory('_data')..createSync();
+      dataDir.childFile('site.json').writeAsStringSync('{"name": "My Site"}');
+      dataDir.childFile('user.yaml').writeAsStringSync('name: Test User');
+      dataDir.childFile('unsupported.md').writeAsStringSync('This should be ignored');
+
+      final loader = FilesystemDataLoader('_data', fileSystem: fileSystem, watcherFactory: (_) => mockWatcher);
+      final page = Page(
+        path: 'test.md',
+        url: '/test',
+        content: '',
+        config: PageConfig(dataLoaders: [loader]),
+        loader: mockLoader,
+      );
+
+      // Act
+      await page.loadData();
+
+      // Assert
+      expect(
+        page.data,
+        equals({
+          'site': {'name': 'My Site'},
+          'user': {'name': 'Test User'},
+          'page': <String, Object?>{},
+        }),
+      );
+    });
+
     test('does nothing if data directory does not exist', () async {
       // Arrange
       final loader = FilesystemDataLoader('_data', fileSystem: fileSystem);
--- a/packages/jaspr_content/test/route_loaders/filesystem_loader_test.dart
+++ b/packages/jaspr_content/test/route_loaders/filesystem_loader_test.dart
@@ -13,16 +13,11 @@ class MockDirectoryWatcher extends Mock implements DirectoryWatcher {}
 class MockPage extends Mock implements Page {}
 
 void main() {
-  late MemoryFileSystem fileSystem;
-
-  setUp(() {
-    fileSystem = MemoryFileSystem();
-  });
-
   group('FilesystemLoader', () {
     group('loadPageSources()', () {
       test('loads basic file structure', () async {
         // Arrange
+        final fileSystem = MemoryFileSystem();
         final pagesDir = fileSystem.directory('pages')..createSync();
         pagesDir.childFile('index.md').createSync();
         pagesDir.childFile('about.html').createSync();
@@ -36,11 +31,15 @@ void main() {
         expect(sources, hasLength(2));
         expect(
           sources,
-          equals([pageSource('index.md', '/', private: false), pageSource('about.html', '/about', private: false)]),
+          equals([
+            pageSource('index.md', '/', private: false),
+            pageSource('about.html', '/about', private: false),
+          ]),
         );
       });
 
       test('loads files with dots in their name', () async {
+        final fileSystem = MemoryFileSystem();
         final pagesDir = fileSystem.directory('pages')..createSync();
         pagesDir.childFile('release-0.1.0.md').createSync();
 
@@ -57,6 +56,7 @@ void main() {
 
       test('loads nested file structure', () async {
         // Arrange
+        final fileSystem = MemoryFileSystem();
         final pagesDir = fileSystem.directory('pages')..createSync();
         pagesDir.childFile('index.md').createSync();
         fileSystem.directory('pages/blog').createSync();
@@ -79,14 +79,14 @@ void main() {
       });
 
       test('loads nested file structure on windows', () async {
-        final MemoryFileSystem fs = MemoryFileSystem(style: FileSystemStyle.windows);
         // Arrange
-        final pagesDir = fs.directory('pages')..createSync();
+        final fileSystem = MemoryFileSystem(style: FileSystemStyle.windows);
+        final pagesDir = fileSystem.directory('pages')..createSync();
         pagesDir.childFile('index.md').createSync();
-        fs.directory(r'pages\blog').createSync();
-        fs.file(r'pages\blog\post-1.md').createSync();
+        fileSystem.directory(r'pages\blog').createSync();
+        fileSystem.file(r'pages\blog\post-1.md').createSync();
 
-        final loader = FilesystemLoader('pages', fileSystem: fs);
+        final loader = FilesystemLoader('pages', fileSystem: fileSystem);
 
         // Act
         final sources = await loader.loadPageSources();
@@ -104,6 +104,7 @@ void main() {
 
       test('marks files and directories starting with an underscore as private', () async {
         // Arrange
+        final fileSystem = MemoryFileSystem();
         final pagesDir = fileSystem.directory('pages')..createSync();
         pagesDir.childFile('index.md').createSync();
         pagesDir.childFile('_config.yaml').createSync();
@@ -129,6 +130,7 @@ void main() {
 
       test('keeps suffix for matching patterns', () async {
         // Arrange
+        final fileSystem = MemoryFileSystem();
         final pagesDir = fileSystem.directory('pages')..createSync();
         pagesDir.childFile('sitemap.xml').createSync();
         pagesDir.childFile('feed.rss').createSync();
@@ -148,6 +150,27 @@ void main() {
           ]),
         );
       });
+
+      test('skips files that do not match the filter extensions', () async {
+        // Arrange
+        final fileSystem = MemoryFileSystem();
+        final pagesDir = fileSystem.directory('pages')..createSync();
+        pagesDir.childFile('index.md').createSync();
+        pagesDir.childFile('about.yaml').createSync();
+
+        final loader = FilesystemLoader(
+          'pages',
+          fileSystem: fileSystem,
+          filterExtensions: {'.md'},
+        );
+
+        // Act
+        final sources = await loader.loadPageSources();
+
+        // Assert
+        expect(sources, hasLength(1));
+        expect(sources, equals([pageSource('index.md', '/', private: false)]));
+      });
     });
 
     group('File Watching', () {
@@ -159,21 +182,37 @@ void main() {
         mockWatcher = MockDirectoryWatcher();
         eventController = StreamController<WatchEvent>.broadcast();
         when(() => mockWatcher.events).thenAnswer((_) => eventController.stream);
-
-        fileSystem.directory('pages').createSync();
-        loader = FilesystemLoader('pages', fileSystem: fileSystem, watcherFactory: (_) => mockWatcher);
       });
 
       test('adds new source on file add event', () async {
         // Arrange
+        final fileSystem = MemoryFileSystem();
+        fileSystem.directory('pages').createSync();
+        loader = FilesystemLoader('pages', fileSystem: fileSystem, watcherFactory: (_) => mockWatcher);
         await loader.loadRoutes((_) => PageConfig(), false);
         expect(loader.sources, isEmpty);
 
         // Act
-        final filePath = fileSystem.path.join('pages', 'about.md');
-        fileSystem.file(filePath).createSync();
+        fileSystem.file('pages/about.md').createSync();
+        eventController.add(WatchEvent(ChangeType.ADD, 'pages/about.md'));
+        await Future<void>.delayed(Duration.zero);
 
-        eventController.add(WatchEvent(ChangeType.ADD, filePath));
+        // Assert
+        expect(loader.sources, hasLength(1));
+        expect(loader.sources.first.url, equals('/about'));
+      });
+
+      test('adds new source on file add event on windows', () async {
+        // Arrange
+        final fileSystem = MemoryFileSystem(style: FileSystemStyle.windows);
+        fileSystem.directory('pages').createSync();
+        loader = FilesystemLoader('pages', fileSystem: fileSystem, watcherFactory: (_) => mockWatcher);
+        await loader.loadRoutes((_) => PageConfig(), false);
+        expect(loader.sources, isEmpty);
+
+        // Act
+        fileSystem.file(r'pages\about.md').createSync();
+        eventController.add(WatchEvent(ChangeType.ADD, r'pages\about.md'));
         await Future<void>.delayed(Duration.zero);
 
         // Assert
@@ -183,14 +222,34 @@ void main() {
 
       test('removes source on file remove event', () async {
         // Arrange
-        final filePath = fileSystem.path.join('pages', 'about.md');
-        fileSystem.file(filePath).createSync();
+        final fileSystem = MemoryFileSystem();
+        fileSystem.directory('pages').createSync();
+        loader = FilesystemLoader('pages', fileSystem: fileSystem, watcherFactory: (_) => mockWatcher);
+        fileSystem.file('pages/about.md').createSync();
 
         await loader.loadRoutes((_) => PageConfig(), false);
         expect(loader.sources, hasLength(1));
 
         // Act
-        eventController.add(WatchEvent(ChangeType.REMOVE, filePath));
+        eventController.add(WatchEvent(ChangeType.REMOVE, 'pages/about.md'));
+        await Future<void>.delayed(Duration.zero);
+
+        // Assert
+        expect(loader.sources, isEmpty);
+      });
+
+      test('removes source on file remove event on windows', () async {
+        // Arrange
+        final fileSystem = MemoryFileSystem(style: FileSystemStyle.windows);
+        fileSystem.directory('pages').createSync();
+        loader = FilesystemLoader('pages', fileSystem: fileSystem, watcherFactory: (_) => mockWatcher);
+        fileSystem.file(r'pages\about.md').createSync();
+
+        await loader.loadRoutes((_) => PageConfig(), false);
+        expect(loader.sources, hasLength(1));
+
+        // Act
+        eventController.add(WatchEvent(ChangeType.REMOVE, r'pages\about.md'));
         await Future<void>.delayed(Duration.zero);
 
         // Assert
@@ -205,10 +264,11 @@ void main() {
         final eventController = StreamController<WatchEvent>.broadcast();
         when(() => mockWatcher.events).thenAnswer((_) => eventController.stream);
 
+        final fileSystem = MemoryFileSystem();
         final pagesDir = fileSystem.directory('pages')..createSync();
         final loader = FilesystemLoader('pages', fileSystem: fileSystem, watcherFactory: (_) => mockWatcher);
 
-        final partialPath = fileSystem.path.join('pages', '_includes', 'header.html');
+        final partialPath = 'pages/_includes/header.html';
         fileSystem.file(partialPath)
           ..createSync(recursive: true)
           ..writeAsStringSync('Old Header');
@@ -241,13 +301,12 @@ void main() {
         final eventController = StreamController<WatchEvent>.broadcast();
         when(() => mockWatcher.events).thenAnswer((_) => eventController.stream);
 
-        final fs = MemoryFileSystem(style: FileSystemStyle.windows);
+        final fileSystem = MemoryFileSystem(style: FileSystemStyle.windows);
+        final pagesDir = fileSystem.directory('pages')..createSync();
+        final loader = FilesystemLoader('pages', fileSystem: fileSystem, watcherFactory: (_) => mockWatcher);
 
-        final pagesDir = fs.directory('pages')..createSync();
-        final loader = FilesystemLoader('pages', fileSystem: fs, watcherFactory: (_) => mockWatcher);
-
-        final partialPath = fs.path.join('pages', '_includes', 'header.html');
-        fs.file(partialPath)
+        final partialPath = r'pages\_includes\header.html';
+        fileSystem.file(partialPath)
           ..createSync(recursive: true)
           ..writeAsStringSync('Old Header');
         pagesDir.childFile('index.md')


schultek:
@tenhobi @parlough Would love to get your feedback on this.

https://docs.jaspr.site/~745/content/concepts/assets

tenhobi (commented):
Looks good and what suprises me is looks quite simple, awesome!

One question, if I create a page using router, such as home page, those routes does not provide page in context (_InheritedPage), and I am not able to access asset manager using context - just maybe by passing reference. But for that I would need to make track which component is passed to page vs using router. Is there a way how to properly setup components from router to be able to also access asset manager and other stuff (i.e. site data from content/_data/site.yml)?
https://docs.jaspr.site/content/concepts/route_loading#adding-other-routes

schultek:
Nice.

Yes you would have to pass down the asset manager yourself or create an inherited component for it.

For accessing page stuff outside, that doesn't really work because there are no pages then. A page aways has some source (e.g. markdown file). But you can use `MemoryLoader` and `MemoryPage.builder` to create a page from your own component. https://docs.jaspr.site/content/concepts/route_loading#memoryloader

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+
+import 'package:crypto/crypto.dart';
+import 'package:jaspr/jaspr.dart';
+import 'package:path/path.dart' as p;
+import 'package:shelf/shelf.dart';
+import 'package:shelf_static/shelf_static.dart';
+
+import '../jaspr_content.dart';
+
+extension ResolveAssetExtension on BuildContext {
```
parlough: Consider moving this extension to the end of file. I think it isn't what contributors or readers are most interested in when opening this file.

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+/// Relative asset paths in the page data (e.g. in frontmatter data) or the page content (e.g. in `<img>` tags)
+/// will be resolved to absolute paths based to be used in the rendered pages.
+///
+/// ## Setup
+///
+/// Create a new [AssetManager] instance in your main method and configure it in your [ContentApp] like this:
+///
+/// ```dart
+/// void main() {
+///
```
parlough:
```suggestion
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+///   final assetManager = AssetManager(...);
+///
+///   // Add middleware to serve assets.
+///   Jaspr.addMiddleware(assetManager.middleware);
```
parlough: It seems this doesn't exist on `Jaspr`, but on `ServerApp`:


```suggestion
///   ServerApp.addMiddleware(assetManager.middleware);
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+///
+///   runApp(ContentApp(
+///     ...,
+///     dataLoaders: [
+///       ...,
+///       // Add data loader to process relative asset paths in page data.
+///       assetManager.dataLoader,
+///     ],
+///     extensions: [
+///       ...,
```
parlough: Consider converting the `...`s into inline comments, so this is easier to copy in case anyone ever does.


```suggestion
///   runApp(ContentApp(
///     // ...
///     dataLoaders: [
///       // ...
///       // Add data loader to process relative asset paths in page data.
///       assetManager.dataLoader,
///     ],
///     extensions: [
///       // ...
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+///       // Add extension to process relative asset paths in page content.
+///       assetManager.pageExtension,
+///     ],
+///   ));
+/// }
+/// ```
+///
+/// ## Usage
+///
+/// There are three ways to relative asset paths are resolved:
```
parlough:
```suggestion
/// There are three ways relative asset paths are resolved:
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+///
+///    For example, when having a `thumbnailImage` property in the frontmatter data of a page, you can add
+///    `'page.thumbnailImage'` to [dataProperties] to resolve any relative paths in this property.
+///
+/// 2. Relative paths in `<img>`, `<video>`, `<audio>` and `<source>` elements in the page content are
+///    automatically resolved.
+///
+///    This also works for images in markdown content.
+///
+/// 3. To resolve relative paths in your own components, use the [resolveAsset] extension method on [BuildContext].
```
parlough: This is resolving to the wrong `resolveAsset`.

```suggestion
/// 3. To resolve relative paths in your own components, use the [ResolveAssetExtension.resolveAsset] extension method on [BuildContext].
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+///      final imagePath = context.resolveAsset(myRelativePath);
+///      return img(src: imagePath);
+///    }
+///    ```
+///
+/// ---
+///
+/// During development, assets are served directly from the [directory].
+///
+/// When building (in static mode), assets are copied to the build/jaspr/[outputPrefix] directory.
```
parlough:
```suggestion
/// When building (in `static` mode), assets are copied to the build/jaspr/[outputPrefix] directory.
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+class AssetManager {
+  static const defaultAssetExtensions = {
```
parlough, line 82: Consider adding a doc comment here:

```suggestion
  /// The default set of file extensions treated as assets.
  ///
  /// This is the default value for [assetExtensions] and includes
  /// common image, video, and audio formats.
  static const defaultAssetExtensions = {
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+  /// the content directory.
+  final String directory;
+
+  /// The prefix to use for the assets paths in the output directory.
+  ///
+  /// Defaults to `assets`. This is added as a prefix to the resolved asset path and
+  /// sets the output directory for the assets relative to the build directory.
+  final String outputPrefix;
+
+  /// The extensions of the assets to be handled.
```
parlough: Consider expanding this doc comment to clarify the expected format (leading dot or not) and to point to the default behavior.

Perhaps:

```suggestion
  /// The file extensions recognized as assets, including
  /// the leading dot (such as `'.png'` for PNG images).
  ///
  /// If not specified, defaults to [defaultAssetExtensions]. 
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+///
+/// When building (in static mode), assets are copied to the build/jaspr/[outputPrefix] directory.
+class AssetManager {
+  static const defaultAssetExtensions = {
+    '.png',
+    '.jpg',
+    '.jpeg',
+    '.gif',
+    '.svg',
+    '.webp',
```
parlough, line 89: Consider adding `.avif` and `.jxl` as they are both becoming widely supported and more common.

```suggestion
    '.jpeg',
    '.jxl',
    '.gif',
    '.svg',
    '.webp',
    '.avif',
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
+import 'dart:io';
```
parlough, line 4: Consider adding a doc import here so the doc comment references to `ServerApp.addMiddleware` can resolve and be linked properly.

```suggestion
/// @docImport 'package:jaspr/server.dart';
library;

import 'dart:io';
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+
+  final Map<(String, Object?), String> _assets = {};
+
+  /// Resolves the asset at [path] relative to the current page. If [path] starts with a slash, it is resolved
+  /// relative to the asset root.
+  ///
+  /// The optional [aspect] parameter can be used to provide additional information to the asset transformers during build.
+  /// Requires an [AssetManager] to be setup.
+  String resolveAsset(String path, Page page, [Object? aspect]) {
+    if (path.startsWith('http')) return path;
```
parlough: Consider being more explicit here, just in case some asset name or path does start with `http` but isn't a link.

Either try parsing it as a URI (`Uri.tryParse`) or check something like `path.startsWith('http://') || path.startsWith('https://')`.

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+
```
parlough, line 192: Since content usually specifies assets in `/` style, should this (and following) logic specifically use `path.posix`?

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+      );
+    }
+
+    return p.normalize(p.join('/', outputPrefix, resolvedPath));
+  }
+}
+
+/// Represents an asset that can be processed by [AssetTransformer].
+sealed class Asset {
+  Asset._();
```
parlough: Since `Asset` is a sealed class (and implicitly abstract), I think you can actually make this just `Asset()` so the children don't need to reference `super._()`.

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+
+  /// Creates an asset from bytes.
```
parlough:
```suggestion
  /// Creates an asset from the specified [bytes].
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+  factory Asset.fromBytes(String path, Uint8List bytes) = MemoryAsset;
+
+  /// Creates an asset from a file.
```
parlough:
```suggestion
  /// Creates an asset from the specified [file].
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+  factory Asset.fromFile(String path, File file) = FileAsset;
+
+  /// The path of the asset.
+  String get path;
+
+  /// Reads the asset as bytes.
+  Uint8List readAsBytes();
+}
+
+/// An asset that is stored in memory.
```
parlough:
```suggestion
/// An asset that is stored in memory as [bytes].
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+
+  @override
+  final String path;
+  final Uint8List bytes;
+
+  @override
+  Uint8List readAsBytes() => bytes;
+}
+
+/// An asset that is stored in a file.
```
parlough:
```suggestion
/// An asset that is stored in a [file].
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+  /// Transforms the given [asset].
+  ///
+  /// The optional [aspect] parameter can be used to provide additional information to the transformer.
+  Asset transform(Asset asset, [Object? aspect]);
+}
+
+/// An asset transformer that adds a hash to the filename of assets.
+///
+/// This is useful for cache busting.
+class HashingAssetTransformer extends AssetTransformer {
```
parlough, line 320: I love it!

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+
+  @override
+  Uint8List readAsBytes() => bytes;
+}
+
+/// An asset that is stored in a file.
+class FileAsset extends Asset {
+  FileAsset(this.path, this.file) : super._() {
+    if (!file.existsSync()) {
+      throw StateError('[ERROR] Asset not found: ${file.path}');
```
parlough:
```suggestion
      throw StateError('Asset file not found: ${file.path}');
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+
+/// An asset transformer that adds a hash to the filename of assets.
+///
+/// This is useful for cache busting.
+class HashingAssetTransformer extends AssetTransformer {
+  const HashingAssetTransformer({this.flattenPath = true});
+
+  /// Whether to flatten the path of the asset.
+  ///
+  /// If true, the path will be flattened to a single level, e.g. `images/foo/bar.png` becomes `bar.<hash>.png`.
```
parlough:
```suggestion
  /// If `true`, the path will be flattened to a single level, e.g. `images/foo/bar.png` becomes `bar.<hash>.png`.
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+  /// If false, the path will be preserved, e.g. `images/foo/bar.png` becomes `images/foo/bar.<hash>.png`.
```
parlough:
```suggestion
  /// If `false`, the path will be preserved, e.g. `images/foo/bar.png` becomes `images/foo/bar.<hash>.png`.
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+        } else {
+          currentData = null;
+          break;
+        }
+      }
+      if (currentData is Map<String, Object?>) {
+        final value = currentData[segments.last];
+        if (value is String) {
+          currentUpdates[segments.last] = assetManager.resolveAsset(value, page);
+        }
```
parlough:
```suggestion
        final lastSegment = segments.last;
        if (currentData[lastSegment] case final String pathValue) {
          currentUpdates[lastSegment] = assetManager.resolveAsset(pathValue, page);
        }
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+}
+
+class _AssetDataLoader implements DataLoader {
+  const _AssetDataLoader(this.assetManager);
+
+  final AssetManager assetManager;
+
+  @override
+  Future<void> loadData(Page page) async {
+    if (assetManager.filterPages != null && !assetManager.filterPages!(page)) {
```
parlough: You have this logic twice and I imagine it will be useful in other cases as well. Perhaps you can add a `shouldProcessPage` function or similar?

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+      'img': 'src',
+      'video': 'src',
+      'audio': 'src',
+      'source': 'src',
+    };
+    return nodes.map((node) {
+      if (node is! ElementNode) return node;
+      final tag = node.tag.toLowerCase();
+      if (targetElements.containsKey(tag)) {
+        final attribute = targetElements[tag]!;
```
parlough:
```suggestion
      if (targetElements[tag] case final attribute?) {
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,398 @@
...
+          }, node.children);
+        }
+      }
+      if (node.children != null) {
+        return ElementNode(
+          node.tag,
+          node.attributes,
+          _processNodes(node.children!, page),
+        );
+      }
```
parlough:
```suggestion
      if (node.children case final nodeChildren?) {
        return ElementNode(
          node.tag,
          node.attributes,
          _processNodes(nodeChildren, page),
        );
      }
```

packages/jaspr_content/test/core/asset_manager_test.dart
```diff
@@ -0,0 +1,175 @@
...
+      });
+    });
+
+    group('resolveAsset', () {
+      test('resolves paths directly', () {
+        when(() => page.path).thenReturn('dir/page.html');
+
+        expect(assetManager.resolveAsset('image.png', page), '/assets/dir/image.png');
+        expect(assetManager.resolveAsset('/absolute.png', page), '/assets/absolute.png');
+        expect(assetManager.resolveAsset('https://example.com/logo.png', page), 'https://example.com/logo.png');
```
parlough, line 172:
```suggestion
        expect(assetManager.resolveAsset('https://example.com/logo.png', page), 'https://example.com/logo.png');
        expect(assetManager.resolveAsset('http://example.com/logo.png', page), 'http://example.com/logo.png');
```

packages/jaspr_content/test/core/asset_manager_test.dart
parlough, line 1: Consider also adding some basic, direct unit tests for `HashingAssetTransformer`.

packages/jaspr/lib/src/server/server_app.dart
```diff
@@ -24,6 +25,15 @@ final class ServerApp {
   static final StreamController<Object> _reassembleController = StreamController.broadcast();
   static Stream<Object> get onReassemble => _reassembleController.stream;
 
+  static final _middleware = <Middleware>[];
```
parlough:
```suggestion
  static final List<Middleware> _middleware = [];
```

packages/jaspr_content/lib/src/route_loader/filesystem_loader.dart
```diff
@@ -28,6 +29,11 @@ class FilesystemLoader extends RouteLoaderBase {
   /// The directory to load pages from.
   final String directory;
 
+  /// A set of file extensions to filter for.
+  ///
+  /// Files in the content directory with other extensions are skipped.
+  final Set<String>? filterExtensions;
```
parlough: Since this is a set already, consider making it non-nullable and adjusting the constructor to default to `const {}`. That way usage of it can be a bit cleaner :) 


```suggestion
  final Set<String> filterExtensions;
```

packages/jaspr_content/CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## Unreleased minor
+
+- New `AssetManager` class for handling content assets. Assets like images, videos, etc. can be co-located with the content and referenced by their relative path from a page.
+  - Assets paths are resolved across markdown files, page data and components using `context.resolveAsset('some/asset.png')`.
```
parlough:
```suggestion
  - Asset paths are resolved across Markdown files, page data and components using `context.resolveAsset('some/asset.png')`.
```

packages/jaspr_content/lib/src/data_loader/data_loader.dart
```diff
@@ -19,6 +19,10 @@ abstract class DataLoader {
   /// The page's data is modified in place.
   Future<void> loadData(Page page);
 
+  static bool canParse(String name) {
```
parlough, line 27:
```suggestion
  /// Whether a file with the specified [name] can be parsed by [DataLoader.parseData].
  ///
  /// Files with other extensions are not recognized as data sources
  /// and will be skipped by the default [FilesystemDataLoader] implementation.
  static bool canParse(String name) {
```

packages/jaspr_content/lib/src/data_loader/data_loader.dart
```diff
@@ -19,6 +19,10 @@ abstract class DataLoader {
...
+    return name.endsWith('.json') || name.endsWith('.yaml') || name.endsWith('.yml');
```
parlough: Optionally consider using `package:path` here:

```suggestion
    return const {'.json', '.yaml', '.yml'}.contains(path.extension(name));
```

docs/content/concepts/assets.mdx
```diff
@@ -0,0 +1,178 @@
...
+1. Create an instance of `AssetManager`.
+2. Add the `assetManager.middleware` to the server app.
+3. Add `assetManager.dataLoader` and `assetManager.pageExtension` to your `ContentApp`.
+
+```dart
+import 'package:jaspr/server.dart';
+import 'package:jaspr_content/jaspr_content.dart';
+
+void main() {
+  // 1. Initialize AssetManager
```
parlough:
```suggestion
  // 1. Initialize and configure AssetManager.
```

docs/content/concepts/assets.mdx
```diff
@@ -0,0 +1,178 @@
...
+    // Resolve 'image.png' relative to the current page.
+    final imagePath = context.resolveAsset('image.png');
+
+    yield img(src: iconPath);
+      
+    // You can also resolve absolute paths (relative to the asset root).
+    final logoPath = context.resolveAsset('/images/logo.png');
+      
+    yield img(src: logoPath);
+  }
```
parlough:
```suggestion
  Component build(BuildContext context) {
    // Resolve 'image.png' relative to the current page.
    final imagePath = context.resolveAsset('image.png');

    // You can also resolve absolute paths (relative to the asset root).
    final logoPath = context.resolveAsset('/images/logo.png');

    return .fragment([
      img(src: iconPath),
      img(src: logoPath),
    ]);
  }
```

docs/content/concepts/assets.mdx
```diff
@@ -0,0 +1,178 @@
...
+You can customize this behavior or add your own transformers by passing a list of `AssetTransformer`s to the `AssetManager` constructor.
+
+```dart
+final assetManager = AssetManager(
+  directory: 'content',
+  assetTransformers: [
+    // Add your custom transformer
+    MyCustomTransformer(),
+    // Use default hashing
+    HashingAssetTransformer(),
```
parlough:
```suggestion
    // Add a custom transformer.
    MyCustomTransformer(),
    // Use the default hashing behavior.
    HashingAssetTransformer(),
```

docs/content/concepts/assets.mdx
```diff
@@ -0,0 +1,178 @@
...
+  @override
+  Asset transform(Asset asset, [Object? aspect]) {
+    // Transform the asset (e.g. compress images, optimize svg, etc.)
+    // Return a new Asset or the same asset.
+    return asset;
+  }
+}
+```
+
+The `aspect` option is passed through as a parameter from `context.resolveAsset(aspect: ...)` and can be used to generate asset variants, e.g. differently resized images.
```
parlough:
```suggestion
The `aspect` option is passed through as a parameter from `context.resolveAsset(aspect: ...)` and can be used to generate asset variants, such as differently resized images.
```

packages/jaspr_content/lib/src/asset_manager.dart
schultek: That's a good catch. I need to separate between url-like paths which are needed in the resulting html, and filepaths, which are needed to read+write the actual assets.

schultek: @parlough Please check the logic again. I should be using p.url for all in-content path logic, and normal `p` for all file paths. Also `p.fromUrl` should give the wanted behavior of transforming an url path to a file path, especially on windows.

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,419 @@
...
+
+    final resolvedPath = path.startsWith('/')
+        ? path.substring(1)
+        : p.url.normalize(p.url.join(p.url.dirname(page.path), path));
+
+    if (kGenerateMode) {
+      final sourcePath = p.normalize(p.join(directory, p.fromUri(resolvedPath)));
+      if (_assets.containsKey((sourcePath, aspect))) {
+        return _assets[(sourcePath, aspect)]!;
+      }
```
parlough: Can avoid the duplicated record creation and map lookup here:

```suggestion
      final assetKey = (sourcePath, aspect);
      if (_assets[assetKey] case final cachedAsset?) {
        return cachedAsset;
      }
```

Then can also use the already built `assetKey` record on L222-223.

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,419 @@
...
+
+      return _assets[(sourcePath, aspect)] = p.url.join('/', outputPrefix, asset.path);
+    }
+
+    return p.url.normalize(p.url.join('/', outputPrefix, resolvedPath));
+  }
+
+  bool _shouldProcessPage(Page page) {
+    return filterPages == null || filterPages!(page);
+  }
```
parlough:
```suggestion
  /// Whether to process the specified [page] with the asset manager.
  bool _shouldProcessPage(Page page) => filterPages?.call(page) ?? true;
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,419 @@
...
+}
+
+extension ResolveAssetExtension on BuildContext {
+  /// Resolves the asset at [path] relative to the current page. If [path] starts with a slash, it is resolved
+  /// relative to the asset root.
```
parlough: Consider explicitly specifying what kind of slash here:

```suggestion
  /// Resolves the asset at [path] relative to the current page.
  ///
  /// If [path] starts with a forward slash (`/`), it is
  /// resolved relative to the asset root.
  ```

File: packages/jaspr_content/lib/src/asset_manager.dart [hunk 40]
```diff
@@ -0,0 +1,419 @@
...
+
+    return p.url.normalize(p.url.join('/', outputPrefix, resolvedPath));
+  }
+
+  bool _shouldProcessPage(Page page) {
+    return filterPages == null || filterPages!(page);
+  }
+}
+
+extension ResolveAssetExtension on BuildContext {
```
parlough, line 236:
```suggestion
/// Provides the [resolveAsset] extension to resolve assets relative to the current page.
extension ResolveAssetExtension on BuildContext {
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,419 @@
...
+        return _assets[(sourcePath, aspect)]!;
+      }
+
+      var asset = Asset.fromFile(resolvedPath, File(sourcePath));
+      for (final transformer in assetTransformers) {
+        asset = transformer.transform(asset, aspect);
+      }
+
+      switch (asset) {
+        case FileAsset(file: final file):
```
parlough: Since the names are the same, can use this shorthand:

```suggestion
        case FileAsset(:final file):
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,419 @@
...
+          final outFile = File(
+            p.join('build', 'jaspr', outputPrefix, p.fromUri(asset.path)),
+          );
+          outFile.createSync(recursive: true);
+          file.copySync(outFile.path);
+        case MemoryAsset(bytes: final bytes):
```
parlough:
```suggestion
        case MemoryAsset(:final bytes):
```

packages/jaspr_content/lib/src/asset_manager.dart
parlough: It's a bit hard to keep track of, but the updated logic looks generally good to me.

I do have one comment/question about one potential issue: https://github.com/schultek/jaspr/pull/745/changes#r2807239365

Also consider adding a inline note/warning comment to the beginning of the function to clarify why and when this distinction is needed here.

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,419 @@
...
+    final maybeAssetManager = page.data[_dataKey];
+    if (maybeAssetManager is! AssetManager) return null;
+    return maybeAssetManager;
+  }
+
+  /// Resolves the asset at [path] relative to the current page. If [path] starts with a slash, it is resolved
+  /// relative to the asset root.
+  ///
+  /// The optional [aspect] parameter can be used to provide additional information to the asset transformers during build.
+  /// Requires an [AssetManager] to be setup.
```
parlough: Since this is a method on `AssetManager`, isn't this already implied? Or what is meant by "to be setup"? Added as a data loader?

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,419 @@
...
+
+  /// Retrieves the [AssetManager] for the given [page].
+  static AssetManager? of(Page page) {
+    final maybeAssetManager = page.data[_dataKey];
+    if (maybeAssetManager is! AssetManager) return null;
+    return maybeAssetManager;
+  }
+
+  /// Resolves the asset at [path] relative to the current page. If [path] starts with a slash, it is resolved
+  /// relative to the asset root.
```
parlough: Consider explicitly specifying what kind of slash here:

```suggestion
  /// Resolves the asset at [path] relative to the current page.
  ///
  /// If [path] starts with a forward slash (`/`), it is
  /// resolved relative to the asset root.
```

packages/jaspr_content/lib/src/asset_manager.dart
```diff
@@ -0,0 +1,419 @@
...
+  ///
+  /// The optional [aspect] parameter can be used to provide additional information to the asset transformers during build.
+  /// Requires an [AssetManager] to be setup.
+  String resolveAsset(String path, Page page, [Object? aspect]) {
+    if (path.startsWith('http://') || path.startsWith('https://')) return path;
+
+    final resolvedPath = path.startsWith('/')
+        ? path.substring(1)
+        : p.url.normalize(p.url.join(p.url.dirname(page.path), path));
```
parlough: Is `page.path` OS-dependent and therefore needs to be converted to url path format here?

Either way, also consider adding to the doc comment of `page.path` to clarify its format.

packages/jaspr_content/lib/src/data_loader/data_loader.dart
```diff
@@ -19,16 +20,22 @@ abstract class DataLoader {
...
   /// Currently supports parsing .json and .yaml data. All other data is returned as is.
   static Object? parseData(String name, String data) {
-    if (name.endsWith('.json')) {
-      return jsonDecode(data);
-    } else if (name.endsWith('.yaml') || name.endsWith('.yml')) {
-      return yaml.loadYamlNode(data).normalize();
-    } else {
-      return data;
-    }
+    return switch (path.extension(name)) {
```
parlough, line 35: Thanks for cleaning this up too :D

packages/jaspr_content/lib/src/asset_manager.dart
schultek: Good catch, this wasn't enforced before.

I changed it so that page.path is always in posix format now.

packages/jaspr_content/CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+## Unreleased minor
```
parlough, line 1: @schultek I think you should consider releasing this as a proper breaking release. I worry the `page.path` change especially might hit people.

```suggestion
## Unreleased breaking
```

schultek: minor **is** breaking for 0-dot versions

PR #759, opened 2026-03-01
[jaspr_cli] Don't ignore errors when web or flutter build fails

parlough:
The `_buildWeb` was already returning an `int` representing its `exitCode` but it was ignored. As far as I can tell, that wasn't intentional as this function is the only caller.


schultek (approved):
Good catch, thanks.

PR #764, opened 2026-03-01
[jaspr/client] Track sibling changes in dom render objects

parlough:
Fixes sibling pointer tracking in DOM render objects when children are removed.

Before this fix, removing a child from a `DomRenderFragment` did not update the `previousSibling`/`nextSibling` linked list or the fragment's `_firstChild`/`_lastChild` pointers, which could cause incorrect DOM ordering when new elements were later inserted near a previously-emptied fragment.

### DIFF
--- a/packages/jaspr/lib/src/client/dom_render_object.dart
+++ b/packages/jaspr/lib/src/client/dom_render_object.dart
@@ -549,27 +549,24 @@ mixin MultiChildDomRenderObject on DomRenderObject {
   void removeChild(DomRenderObject child) {
     if (child case DomRenderFragment(isAttached: true)) {
       child.removeChildren(this);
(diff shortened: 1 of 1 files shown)


schultek:
Looks like a good fix, but I wonder how did you discover this? A bug you ran into?

parlough:
> Looks like a good fix, but I wonder how did you discover this? A bug you ran into?

I'm exploring building a UI primitive library and when having AI write tests for it, they hit this issue. So not in an actual app. Then I had the AI find the potential sources for the issue as well :)

I updated the description with some additional changes of what the PR does as well. I also pushed an additional commit to avoid duplicating the new unlinking logic, so one of them isn't accidentally updated without the other.

PR #770, opened 2026-03-05
[jaspr_content] Avoid file descriptor exhaustion during an eager load

parlough:
Some contributors were hitting this on macOS when building docs.flutter.dev, since the load call ran until it awaited I/O calls. Then by the time we `await`ed the results of the loads, potentially many file descriptors were open.

To avoid this case, when eagerly loading page sources:

- Don't call `load` on them right away, just track the sources themselves.
- After each source is tracked, load them in batches.

### DIFF
--- a/packages/jaspr_content/lib/src/route_loader/route_loader.dart
+++ b/packages/jaspr_content/lib/src/route_loader/route_loader.dart
@@ -131,7 +131,7 @@ abstract class RouteLoaderBase<T extends PageSource> implements RouteLoader {
 
     // Use a pool to avoid running out of file descriptors.
     if (eagerSources.isNotEmpty) {
-      const maxConcurrentLoads = 32;
+      const maxConcurrentLoads = 64;
       final pool = Pool(maxConcurrentLoads);
       await eagerSources.map((s) => pool.withResource(s.load)).wait;
       await pool.close();
--- a/packages/jaspr_content/test/route_loaders/route_loader_test.dart
+++ b/packages/jaspr_content/test/route_loaders/route_loader_test.dart
@@ -14,7 +14,7 @@ void main() {
 
       expect(routes, hasLength(sourceCount));
       // Should match the maximum defined in the implementation.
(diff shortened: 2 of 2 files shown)


parlough:
@schultek Any thoughts on this solution? At first I considered adjusting the load logic itself, but this seemed more flexible as weren't not in control of the `load` logic. It could come from a third-party `PageSource` implementation.

packages/jaspr_content/lib/src/route_loader/route_loader.dart
```diff
@@ -129,6 +128,14 @@ abstract class RouteLoaderBase<T extends PageSource> implements RouteLoader {
       }
     }
 
+    // Load page sources in batches to avoid running out of file descriptors.
+    if (eagerSources.isNotEmpty) {
+      const maxConcurrentLoads = 32;
```
schultek: I think we can increase this to at least 64, from what I read the default limit is 256

packages/jaspr_content/lib/src/route_loader/route_loader.dart
```diff
@@ -129,6 +128,14 @@ abstract class RouteLoaderBase<T extends PageSource> implements RouteLoader {
...
+      for (final batch in eagerSources.slices(maxConcurrentLoads)) {
```
schultek: We could improve this by having a queue of running futures and load a new one every time a future completes. 
Right now if one of the batch loads is slow this blocks all other loads of that batch.

parlough: Good point. Rather than rolling our own logic here, I went ahead and used [`package:pool`](https://pub.dev/packages/pool) from the Dart team. Does adding that dependency seem fine to you or would you prefer we roll our own solution here?

schultek: TIL. Fine with me to add the dependency.

PR #776, opened 2026-03-16
Server Components

schultek:
## Description

Add support for **Server Components** to Jaspr.

## Type of Change

- ❌ Breaking change


packages/jaspr/lib/src/client/component_anchors.dart
```diff
@@ -6,47 +6,150 @@ import 'package:universal_web/web.dart' as web;
...
 
   late final Map<String, Object?> params = data != null
       ? jsonDecode(const DomValidator().unescapeMarkerText(data!)) as Map<String, Object?>
       : {};
 
   Future<void> resolve() async {
-    final r = await builder;
-    builder = r;
+    final r = await (Future.value(builder), Future.wait(serverAnchors.values.map((a) => a.resolve()))).wait;
+    builder = r.$1;
```
parlough:
```suggestion
    final (clientBuilder, _) = await (
      Future.value(builder),
      serverAnchors.values.map((a) => a.resolve()).wait,
    ).wait;
    builder = clientBuilder;
```

packages/jaspr/lib/src/client/component_anchors.dart
```diff
@@ -6,47 +6,150 @@ import 'package:universal_web/web.dart' as web;
...
 }
 
-final _clientStartRegex = RegExp('^${DomValidator.clientMarkerPrefixRegex}(\\S+)(?:\\s+data=(.*))?\$');
+class ServerComponentAnchor extends ComponentAnchor {
+  ServerComponentAnchor(super.name, super.key, super.startNode);
+
+  final List<ClientComponentAnchor> clientAnchors = [];
+
+  Future<void> resolve() async {
+    await Future.wait(clientAnchors.map((a) => a.resolve()));
```
parlough:
```suggestion
    await [
      for (final anchor in clientAnchors) anchor.resolve(),
    ].wait;
```

packages/jaspr/lib/src/client/options.dart
```diff
@@ -67,4 +69,28 @@ final class ClientLoader {
 ///
 /// **DO NOT USE DIRECTLY.**
 /// Use the generated `defaultClientOptions` instead.
-typedef ClientBuilder = Component Function(Map<String, dynamic> params);
+typedef ClientBuilder = Component Function(ClientParams params);
+
+class ClientParams {
+  ClientParams(this._params, this.serverComponents);
+
+  final Map<String, dynamic> _params;
```
parlough:
```suggestion
  final Map<String, Object?> _params;
```

packages/jaspr/lib/src/client/options.dart
```diff
@@ -67,4 +69,28 @@ final class ClientLoader {
 ///
 /// **DO NOT USE DIRECTLY.**
 /// Use the generated `defaultClientOptions` instead.
-typedef ClientBuilder = Component Function(Map<String, dynamic> params);
+typedef ClientBuilder = Component Function(ClientParams params);
+
+class ClientParams {
```
parlough:
```suggestion
final class ClientParams {
```

packages/jaspr/lib/src/client/options.dart
```diff
@@ -67,4 +69,28 @@ final class ClientLoader {
...
+    } else {
+      return const Component.text('');
+    }
+  }
+
+  T get<T>(String key) {
+    if (_params[key] is! T) {
+      print('$key is not $T: ${_params[key]}');
+    }
+    return _params[key] as T;
```
parlough: To avoid the duplicate map lookup:

```suggestion
    final value = _params[key];
    if (value is! T) {
      throw StateError('$key is not $T: $value');
    }

    return value;
```

packages/jaspr/lib/src/client/slotted_child_view.dart
parlough: Can this be made non-null? The following code assumes it is now.

```suggestion
  factory SlottedDomRenderObject.fromNodes(List<web.Node>? nodes, DomRenderObject parent) {
```

packages/jaspr/lib/src/client/run_app.dart
```diff
@@ -3,5 +3,7 @@ import 'client_binding.dart';
 
 /// Main entry point for the browser app.
 ClientAppBinding runApp(Component app, {String attachTo = 'body'}) {
-  return ClientAppBinding()..attachRootComponent(app, attachTo: attachTo);
+  final binding = ClientAppBinding();
+  Future.microtask(() => binding.attachRootComponent(app, attachTo: attachTo));
```
parlough: Can you add an inline comment explaining why it's being scheduled as a microtask? Is it so it runs before handling any SSE events?

packages/jaspr/lib/src/server/adapters/client_component_adapter.dart
```diff
@@ -1,81 +1,100 @@
...
+  @override
+  void prepareBoundary(ChildListRange range) {
+    element.visitAncestorElements((e) {
+      if (registry.serverElements.containsValue(e)) {
+        return false;
+      } else if (registry.clientElements.containsKey(e)) {
+        _isClientBoundary = false;
+        return false;
+      }
+      return true;
```
parlough:
```suggestion
      if (registry.clientElements.containsKey(e)) {
        _isClientBoundary = false;
        return false;
      }

      if (registry.serverElements.containsValue(e)) {
        return false;
      }

      return true;
```

packages/jaspr/lib/src/client/client_binding.dart
```diff
@@ -66,4 +69,61 @@ class ClientAppBinding extends AppBinding with ComponentsBinding {
   void reportBuildError(Element element, Object error, StackTrace stackTrace) {
     web.console.error('Error while building ${element.component.runtimeType}:\n$error\n\n$stackTrace'.toJS);
   }
+
+  SseClient? _eventsClient;
+
+  void initializeEvents() {
+    if (_eventsClient != null) return;
+
+    _eventsClient = SseClient(r'/$jasprEventsHandler');
```
parlough: Should this always be setup? Or only when debugging?

packages/jaspr/lib/src/framework/components_binding.dart
```diff
@@ -41,6 +41,13 @@ mixin ComponentsBinding on AppBinding {
   }
 }
 
+@visibleForTesting
+class GlobalComponentsBinding {
```
parlough: If this is just for namespacing the static method, then consider making this `abstract final`.

```suggestion
abstract final class GlobalComponentsBinding {
```

packages/jaspr/lib/src/server/adapters/element_boundary_adapter.dart
```diff
@@ -8,16 +8,17 @@ import '../server_binding.dart';
 export '../child_nodes.dart' show ChildListRange, ChildNodeData;
 
 abstract class ElementBoundaryAdapter extends RenderAdapter {
-  ElementBoundaryAdapter(this.element);
+  ElementBoundaryAdapter(this.element, {this.priority = 0});
 
   final Element element;
+  final int priority;
```
parlough, line 19: Can you add a doc comment to this new `priority` field? It's hard to understand what its used for when viewing it insolation.

packages/jaspr/lib/src/server/adapters/document_structure_helper.dart
```diff
@@ -0,0 +1,49 @@
+import '../markup_render_object.dart';
+
+(MarkupRenderElement, MarkupRenderElement, MarkupRenderElement) createDocumentStructure(
```
parlough: Consider using named fields here, otherwise it's hard for the caller (even if it's ourselves) to know what each representation without reading through the implementation.

packages/jaspr/lib/src/server/adapters/document_structure_helper.dart
```diff
@@ -0,0 +1,49 @@
...
+  MarkupRenderObject root, [
+  bool includeDoctype = false,
+]) {
+  var html = root.children.findWhere<MarkupRenderElement>((c) => c.tag == 'html')?.node as MarkupRenderElement?;
+  if (html == null) {
+    final range = root.children.range();
+    root.children.insertAfter(html = root.createChildRenderElement('html')..children.insertNodeAfter(range));
```
parlough: Consider splitting these two make it a bit easier to read:

```suggestion
    html = root.createChildRenderElement('html')..children.insertNodeAfter(range);
    root.children.insertAfter(html);
```

packages/jaspr/lib/src/server/components/client_component_registry.dart
```diff
@@ -0,0 +1,52 @@
...
+    final binding = this.binding as ServerAppBinding;
+
+    final entry = binding.options.clients?[element.component.runtimeType];
+    if (entry == null) {
+      return;
+    }
+
+    if (!_didHandleClientScript) {
+      if (binding.options.clientId != null) {
+        (binding).addRenderAdapter(ClientScriptAdapter(binding.options.clientId!));
```
parlough:
```suggestion
      if (binding.options.clientId case final clientId?) {
        binding.addRenderAdapter(ClientScriptAdapter(clientId));
```

packages/jaspr/lib/src/server/components/client_component_registry.dart
```diff
@@ -0,0 +1,52 @@
...
+      } else if (kDebugMode) {
+        (binding).addRenderAdapter(NoClientScriptAdapter());
```
parlough:
```suggestion
        binding.addRenderAdapter(NoClientScriptAdapter());
```

packages/jaspr/lib/src/server/server_binding.dart
```diff
@@ -66,23 +68,36 @@ class ServerAppBinding extends AppBinding with ComponentsBinding {
     }
 
     final root = rootElement.renderObject as MarkupRenderObject;
-    final adapters = [..._adapters.reversed, GlobalStylesAdapter(this), if (!standalone) DocumentAdapter()];
 
-    for (final adapter in adapters.reversed) {
-      final r = adapter.prepare();
+    for (var i = 0; i < _adapters.length; i++) {
+      final r = _adapters[i].prepare();
```
parlough, line 77: Any reason for this to be an indexed loop?

```suggestion
    for (final adapter in _adapters) {
      final r = adapter.prepare();
```

packages/jaspr/lib/src/server/server_binding.dart
```diff
@@ -66,23 +68,36 @@ class ServerAppBinding extends AppBinding with ComponentsBinding {
...
 
+    if (!standalone) {
+      createDocumentStructure(root, true);
+    }
+
     if (_responseBodyOverride case final override?) {
       return override;
     }
 
+    if (request.headers.singleValues['X-Jaspr-Reload'] == 'true' && !standalone) {
```
parlough:
```suggestion
    if (!standalone && request.headers.singleValues['X-Jaspr-Reload'] == 'true') {
```

packages/jaspr/lib/src/client/client_binding.dart
```diff
@@ -66,4 +69,61 @@ class ClientAppBinding extends AppBinding with ComponentsBinding {
...
+      return;
+    }
+
+    final sw = Stopwatch()..start();
+
+    final newNode = _attachTarget == 'body' ? body : body.querySelector(_attachTarget)!;
+
+    final rootRenderObject = rootElement!.renderObject as RootDomRenderObject;
+    rootRenderObject.node = newNode;
+    rootRenderObject.toHydrate = [...newNode.childNodes.toIterable()];
```
parlough: Maybe encapsulate this logic in a new method on `RootDomRenderObject`, so `node` doesn't need to have a public setter.

packages/jaspr/lib/src/stub/app_context_stub.dart
```diff
@@ -8,4 +8,9 @@ extension AppContext on BuildContext {
   String get url {
     throw UnimplementedError('Should be overridden on client and server.');
   }
+
+  /// Reloads the current page.
+  void reload([String? path]) {
+    throw UnimplementedError('Should be overridden on client and server.');
```
parlough: It seems the client's `AppContext` extension doesn't have this method. Is that intended?

packages/jaspr_builder/lib/src/codec/codecs.dart
```diff
@@ -75,56 +75,80 @@ class EncoderVisitor extends UnifyingTypeVisitorWithArgument<String?, String> {
...
+
+      final decoder = type.nullabilitySuffix == NullabilitySuffix.question
+          ? '$argRaw != null ? $decoderCall($argument) : null'
+          : '$decoderCall($argument)';
+
+      return (decoder: decoder, cast: codec.rawType);
+    } else if (allowComponent &&
+        type.element.name == 'Component' &&
+        type.element.library.identifier == 'package:jaspr/src/framework/framework.dart') {
+      return (decoder: 'p.mount($argument)', cast: 'String');
```
parlough, line 154: What if the component parameter is nullable? Should this consider the nullability suffix like the above handling?

parlough (commented):
Exciting!! I haven't tried it out yet, but some small questions and a few minor suggestions to consider:

packages/jaspr/lib/src/client/client_binding.dart
schultek: True only in debug mode

packages/jaspr/lib/src/server/server_binding.dart
schultek: Can't use foreach because new adapters can be added during .prepare(). I will add a comment.

packages/jaspr/lib/src/stub/app_context_stub.dart
schultek: Ah no, that is leftover from I feature I was experimenting with but decided against for the initial version of server components. Will remove.

packages/jaspr/lib/src/client/slotted_child_view.dart
schultek: True

packages/jaspr/lib/src/client/run_app.dart
schultek: Oh no I forgot why I changed this. I also can't think of a reason right now why it would be needed. Need to think about it.

packages/jaspr/lib/src/server/adapters/client_component_adapter.dart
schultek: Why change the order?

packages/jaspr/lib/src/server/adapters/element_boundary_adapter.dart
schultek: Done

packages/jaspr/lib/src/client/client_binding.dart
schultek: Done. Also I noticed I could remove some dead code in RootDomRenderObject.

packages/jaspr_builder/lib/src/codec/codecs.dart
schultek: Good catch

packages/jaspr/lib/src/client/slotted_child_view.dart
```diff
@@ -120,18 +123,179 @@ class SlottedChildViewElement extends MultiChildRenderObjectElement {
...
+  final Map<web.HTMLElement, ApplyParams?> _appliedParams = {};
+  final Map<web.HTMLElement, Map<String, EventBinding>?> _appliedEvents = {};
+
   @override
-  void updateRenderObject(RenderObject renderObject) {}
+  void updateRenderObject(SlottedDomRenderObject renderObject) {
+    for (final element in _appliedParams.keys) {
+      resetElementParams(element);
+    }
+    _appliedParams.clear();
```
parlough, line 151: Should `_appliedEvents` be cleared here as well?

packages/jaspr/lib/src/framework/framework.dart
```diff
@@ -136,20 +137,20 @@ abstract class Component {
...
   /// ```html
-  /// <div class="wrapping-class some-class" style="padding: 8px; background-color: red;">
+  /// <div class="applied-class some-class" style="padding: 8px; background-color: red;">
   ///   Hello World
   /// </div>
   /// ```
   ///
-  const factory Component.wrapElement({
-    Key? key,
+  const factory Component.apply({
```
parlough, line 145: Should the `ComponentFactoryMigration` be updated to use this new factory instead?

https://github.com/schultek/jaspr/blob/1ad049e511e4c7eb223ef1d2d08ecc071e20e84c/packages/jaspr_cli/lib/src/migrations/component_factory_migration.dart#L71-L77

docs/going_further/server_components.mdx
```diff
@@ -0,0 +1,233 @@
...
+```
+
+Here is how you use the `ExpandableCard` component on the server, passing in a server-only component as a parameter:
+
+```dart title="lib/main.server.dart"
+import 'package:jaspr/server.dart';
+import 'components/expandable_card.dart';
+
+void main() {
+  Jaspr.initializeAll(options: defaultServerOptions);
```
parlough: Should this be `initializeApp`?

docs/api/utils/at_client.mdx
parlough: This prose needs a slight update to account for server components in the sub tree.

docs/going_further/server_components.mdx
```diff
@@ -0,0 +1,233 @@
+---
+title: "Server Components"
+description: Learn how to pass server-side components into client-side interactive boundaries to build high-performance layouts.
+---
+
+When building modern web applications, we want both rich, fluid interactivity and fast initial page loads. 
+
+Typically, once you make a parent component interactive by decorating it with `@client`, Jaspr must compile it and **all of its nested children** to JavaScript so they can run in the browser. Without Server Components, the moment a layout or parent container needs some client-side interactivity, every single child in its subtree gets pulled into the client-side JavaScript bundle—even if those children are completely static.
```
parlough: Consider adjusting this to account for possibility of compiling to Wasm.

```suggestion
Typically, once you make a parent component interactive by decorating it with `@client`, Jaspr must compile it and **all of its nested children** so they can run in the browser. Without Server Components, the moment a layout or parent container needs some client-side interactivity, every single child in its subtree gets pulled into the client-side bundle—even if those children are completely static.
```

docs/going_further/server_components.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+**Server Components** solve this by enabling **surgical hydration**. They allow you to render server-side-only components inside parts of an interactive, hydrated client-side application. With this pattern, you can build interactive client-side wrappers (like expandable cards, navigation shells, sidebars, or tabs) while completely skipping compiling or shipping the static or heavy content inside them.
+
+As a result:
+- **Reduced Bundle Size:** The Dart code for server-only components is never compiled to JavaScript or shipped to the browser, significantly reducing the application's JavaScript footprint.
```
parlough:
```suggestion
- **Reduced Bundle Size:** The Dart code for server-only components is never compiled to JavaScript/Wasm or shipped to the browser, significantly reducing the application's client-side footprint.
```

docs/going_further/server_components.mdx
```diff
@@ -0,0 +1,233 @@
...
+- **Surgical Interactivity:** You only ship the parts of the code to the client that actually need to be interactive, skipping parts that are static.
+- **Faster Page Loading:** Less JavaScript to download, parse, and execute means much faster Time-to-Interactive (TTI) for your users.
```
parlough:
```suggestion
- **Faster Page Loading:** Less code to download, parse, and execute means much faster Time-to-Interactive (TTI) for your users.
```

parlough: Not that important, but it seems TTI was deprecated as a metric, at least for Chrome and Lighthouse. Perhaps a modern equivalent that is improved can be referenced and cross linked to?

TTI deprecation refernece: https://developer.chrome.com/docs/lighthouse/performance/interactive

docs/going_further/server_components.mdx
```diff
@@ -0,0 +1,233 @@
...
+- **Access to Server Resources:** The server components within the client wrapper can safely fetch database records or read files using native APIs like `dart:io`.
+
+## How to use Server Component
+
+It is important to understand that **Server Components are not a separate API** in Jaspr. There is no special annotation, and you don't need to inherit from a special class.
+
+Instead, Server Components are a design paradigm enabled by Jaspr's compilation and hydration model. You achieve this simply by **passing components as parameters** (such as a `child` component, or lists/maps of components) to a component annotated with [`@client`](/api/utils/at_client).
+
+## A Simple Example: Expandable Card
```
parlough, line 24: I like the idea of building something to show the feature. It works well here and is the easiest way to build an understanding of it.

docs/going_further/server_components.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+Let's build a client-side component called `ExpandableCard`. It features a button that lets users toggle the visibility of the card's content. To do this, the card must be interactive and annotated with `@client`. However, the content inside the card is static content that doesn't need client-side interactivity.
+
+Here is the interactive card component:
+
+```dart title="lib/components/expandable_card.dart"
+import 'package:jaspr/jaspr.dart';
```
parlough, line 32:
```suggestion
import 'package:jaspr/dom.dart';
import 'package:jaspr/jaspr.dart';
```

docs/going_further/server_components.mdx
```diff
@@ -0,0 +1,233 @@
...
+  bool showContent = false; // Evaluates to false on the server
+
+  @override
+  Component build(BuildContext context) {
+    return div([
+      button(onClick: () => setState(() => showContent = true), [Text('Reveal')]),
+      // WARNING: If showContent is false during SSR, 'widget.content' is not built on the server.
+      // When showContent becomes true on the client, it will render as EMPTY because 
+      // the client does not have the Dart code to construct the server component.
+      if (showContent) widget.content,
```
parlough:
```suggestion
      // WARNING: If showContent is false during SSR, 'component.content' is not built on the server.
      // When showContent becomes true on the client, it will render as EMPTY because 
      // the client does not have the Dart code to construct the server component.
      if (showContent) component.content, 
```

docs/going_further/server_components.mdx
```diff
@@ -0,0 +1,233 @@
...
+  State<ConditionalCard> createState() => _ConditionalCardState();
+}
+
+class _ConditionalCardState extends State<ConditionalCard> {
+  bool showContent = false; // Evaluates to false on the server
+
+  @override
+  Component build(BuildContext context) {
+    return div([
+      button(onClick: () => setState(() => showContent = true), [Text('Reveal')]),
```
parlough:
```suggestion
      button(onClick: () => setState(() => showContent = true), [.text('Reveal')]),
```

packages/jaspr/lib/src/server/adapters/client_component_adapter.dart
```diff
@@ -1,81 +1,100 @@
...
+        return o;
+      },
+    );
+    return data;
+  }
 
-  @override
-  void willRebuildElement(Element element) {}
+  String getDataForServerComponent(Component component, Element parent) {
+    if (registry.serverComponents[component] case final s?) {
```
parlough: Is this safe? Say if two different elements on the page are from the same component but with different data? Do they need their own marker?

parlough (dismissed):
Still love it. Some small fixes, a few suggestions, and a question but otherwise I think it's in a great place for developers to start trying it out.

packages/jaspr/lib/src/framework/framework.dart
schultek: Updated, and added a new migration for newer versions

docs/api/utils/at_client.mdx
schultek: Done

docs/going_further/server_components.mdx
schultek: Fixed

docs/going_further/server_components.mdx
schultek: Done

docs/going_further/server_components.mdx
schultek: Done

docs/going_further/server_components.mdx
schultek: Done

packages/jaspr/lib/src/server/adapters/client_component_adapter.dart
schultek: Fixed the case that a component is used multiple times.

docs/going_further/server_components.mdx
schultek: reworded

packages/jaspr/lib/src/framework/framework.dart
```diff
@@ -136,20 +137,20 @@ abstract class Component {
...
+    ApplyTarget target,
     String? id,
```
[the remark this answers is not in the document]
schultek: Wrong, since it's a redirecting contructor

PR #777, opened 2026-03-16
Feat: Component.apply with target

schultek:
## Description

Rename `Component.wrapElement` to `Component.apply` and add `ApplyTarget target` parameter

## Type of Change

- ❌ Breaking change

## Ready Checklist

### DIFF
--- a/packages/jaspr/lib/src/framework/components.dart
+++ b/packages/jaspr/lib/src/framework/components.dart
@@ -36,22 +36,32 @@ class DomElement extends MultiChildRenderObjectElement {
   @override
   DomComponent get component => super.component as DomComponent;
 
-  InheritedElement? _wrappingElement;
+  InheritedElement? _inheritedDomElement;
 
   @override
-  List<Component> buildChildren() => component.children ?? [];
+  List<Component> buildChildren() {
+    if (_inheritedDomElement case final inheritedDomElement?) {
+      final inheritedDomComponent = dependOnInheritedElement(inheritedDomElement) as _InheritedDomComponent;
+      return _InheritedDomComponent.consume(
+        inheritedDomComponent.paramsBySelector,
+        component,
+        component.children ?? const [],
+      );
+    }
+    return component.children ?? const [];
+  }
 
   @override
   void _updateInheritance() {
     super._updateInheritance();
     if (_inheritedElements case final originalInheritedElements?
-        when originalInheritedElements.containsKey(_WrappingDomComponent)) {
+        when originalInheritedElements.containsKey(_InheritedDomComponent)) {
       final updatedInheritedElements = HashMap<Type, InheritedElement>.of(originalInheritedElements);
-      _wrappingElement = updatedInheritedElements.remove(_WrappingDomComponent);
+      _inheritedDomElement = updatedInheritedElements.remove(_InheritedDomComponent);
       _inheritedElements = updatedInheritedElements;
       return;
     }
-    _wrappingElement = null;
+    _inheritedDomElement = null;
   }
 
   @override
@@ -85,15 +95,18 @@ class DomElement extends MultiChildRenderObjectElement {
 
   @override
   void updateRenderObject(RenderElement renderObject) {
-    if (_wrappingElement != null) {
-      final wrappingComponent = dependOnInheritedElement(_wrappingElement!) as _WrappingDomComponent;
-      renderObject.update(
-        component.id ?? wrappingComponent.id,
-        _joinString(wrappingComponent.classes, component.classes),
-        _joinMap(wrappingComponent.styles?.properties, component.styles?.properties),
-        _joinMap(wrappingComponent.attributes, component.attributes),
-        _joinMap(wrappingComponent.events, component.events),
-      );
+    if (_inheritedDomElement != null) {
+      final inheritedComponent = dependOnInheritedElement(_inheritedDomElement!) as _InheritedDomComponent;
+      final params = _InheritedDomComponent.applyTo(component, inheritedComponent.paramsBySelector);
+      if (params != null) {
+        renderObject.update(
+          component.id ?? params.id,
+          _DomParams._joinString(params.classes, component.classes),
+          _DomParams._joinMap(params.styles?.properties, component.styles?.properties),
+          _DomParams._joinMap(params.attributes, component.attributes),
+          _DomParams._joinMap(params.events, component.events),
+        );
+      }
 
       return;
     }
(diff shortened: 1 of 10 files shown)


packages/jaspr/lib/src/framework/components.dart
```diff
@@ -144,16 +156,176 @@ class _WrappingDomComponent extends InheritedComponent implements DomComponent {
...
+      ),
+      child: child,
+    );
+  }
+}
+
+/// Defines the target for a [Component.apply] component.
+///
+/// The target can select either a direct child or descendant, and optionally filter by tag, id, and classes.
+class ApplyTarget {
```
parlough:
```suggestion
final class ApplyTarget {
```

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -144,16 +156,176 @@ class _WrappingDomComponent extends InheritedComponent implements DomComponent {
...
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+    return other is ApplyTarget &&
+        onlyChildren == other.onlyChildren &&
+        tag == other.tag &&
+        id == other.id &&
+        classes?.length == other.classes?.length &&
+        classes?.every((c) => other.classes?.contains(c) ?? false) == true;
```
parlough: I believe this doesn't account for the case of them both having `null` for `classes`.

Consider extracting this out to a utility function for comparing sets, like Flutter has with https://api.flutter.dev/flutter/foundation/setEquals.html .

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -85,52 +72,77 @@ class DomElement extends MultiChildRenderObjectElement {
...
+    var classes = component.classes;
+    var styles = component.styles?.properties;
+    var attributes = component.attributes;
+    var events = component.events;
+
+    if (inheritedDomParams case final inheritedDomParams?) {
+      final classesSet = classes?.split(' ').toSet() ?? {};
+      styles = Map.of(styles ?? {});
+      attributes = Map.of(attributes ?? {});
+      events = Map.of(events ?? {});
```
parlough, line 85: To avoid remaking an empty map:

```suggestion
      styles = {...?styles};
      attributes = {...?attributes};
      events = {...?events};
```

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -85,52 +72,77 @@ class DomElement extends MultiChildRenderObjectElement {
...
+
+        if (param.classes case final paramClasses?) {
+          classesSet.addAll(paramClasses);
+        }
+
+        if (param.styles case final stylesProps?) {
+          for (final MapEntry(:key, :value) in stylesProps.entries) {
+            if (!styles.containsKey(key)) {
+              styles[key] = value;
+            }
```
parlough: Consider:

```suggestion
            styles[key] ??= value;
```

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -85,52 +72,77 @@ class DomElement extends MultiChildRenderObjectElement {
...
+          }
+        }
+
+        if (param.attributes case final attributesMap?) {
+          for (final MapEntry(:key, :value) in attributesMap.entries) {
+            if (!attributes.containsKey(key)) {
+              attributes[key] = value;
+            }
```
parlough: Consider:

```suggestion
            attributes[key] ??= value;
```

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -85,52 +72,77 @@ class DomElement extends MultiChildRenderObjectElement {
...
+          }
+        }
+
+        if (param.events case final eventsMap?) {
+          for (final MapEntry(:key, :value) in eventsMap.entries) {
+            if (!events.containsKey(key)) {
+              events[key] = value;
+            }
```
parlough: Consider:

```suggestion
            events[key] ??= value;
```

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -144,16 +156,176 @@ class _WrappingDomComponent extends InheritedComponent implements DomComponent {
...
+        events: events,
+      ),
+      child: child,
+    );
+  }
+}
+
+/// Defines the target for a [Component.apply] component.
+///
+/// The target can select either a direct child or descendant, and optionally filter by tag, id, and classes.
```
parlough:
```suggestion
/// The target can select either a direct child or descendant,
/// and optionally filter by [tag], [id], and [classes].
```

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -144,16 +156,176 @@ class _WrappingDomComponent extends InheritedComponent implements DomComponent {
...
+  final Set<String>? classes;
+
+  String get query => [
+    if (onlyChildren) '> ',
+    if (tag == null && id == null && classes == null)
+      '*'
+    else ...[
+      if (tag != null) tag,
+      if (id != null) '#$id',
+      if (classes != null) ...classes!.map((c) => '.$c'),
```
parlough:
```suggestion
      ?tag,
      if (id != null) '#$id',
      if (classes case final classes?)
        for (final c in classes) '.$c',
```

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -144,16 +156,176 @@ class _WrappingDomComponent extends InheritedComponent implements DomComponent {
...
+    if (current == null) {
+      return _InheritedDomComponent(params: [params], child: child);
+    }
+    return _InheritedDomComponent(params: [...current.params, params], child: child);
+  }
+
+  @override
+  bool updateShouldNotify(_InheritedDomComponent oldComponent) {
+    if (params.length != oldComponent.params.length) return true;
+    return params.indexed.any((e) => oldComponent.params[e.$1] != e.$2);
```
parlough: Consider replacing this with a normal loop. Reads better (to me) than the indexed record accesses and avoids the additional iterator and callback cost.

```suggestion
    for (var i = 0; i < params.length; i += 1) {
      final param = params[i];
      if (oldComponent.params[i] != param) return true;
    }
    return false;
```

parlough: Currently `WrapParams` doesn't have custom equality so this is comparing by identity. Is that intentional/desired?

packages/jaspr/lib/src/framework/framework.dart
```diff
@@ -136,20 +137,20 @@ abstract class Component {
...
   /// ```html
-  /// <div class="wrapping-class some-class" style="padding: 8px; background-color: red;">
+  /// <div class="applied-class some-class" style="padding: 8px; background-color: red;">
   ///   Hello World
   /// </div>
   /// ```
   ///
-  const factory Component.wrapElement({
-    Key? key,
+  const factory Component.apply({
```
parlough, line 145: Apply does make more sense as a name 👍

parlough (commented):
Cool improvement, thanks! I haven't tried it yet, but overall looks good to me.

I will make time to try it out tomorrow.

packages/jaspr/lib/src/framework/components.dart
schultek: I wonder if there is a performance difference between `Map.of` and spread operator.

packages/jaspr/lib/src/framework/components.dart
schultek: Good idea

packages/jaspr/lib/src/framework/components.dart
schultek: True. I don't think it's worth it performance wise to do a deep equality on WrapParams since it can get quite large with styles and attributes etc.. But then actually shallow equality also doesn't make sense because it will always be false since a new instance is created on each build of _ApplyDomComponent.

So actually we can just return true here. Then it's up to the user to not cause too many rebuilds of .apply components when no parameters change.

schultek: But then I can also remove the equality override for ApplyTarget.

packages/jaspr/lib/src/framework/components.dart
schultek: Actually I remove the equality override, see comment below.

PR #815, opened 2026-05-29
Enable DDC's Library Bundle module system by default

Markzipan:
We're updating webdev to compile DDC apps using an updated Library Bundle module format by default, replacing the legacy AMD/Require module format. This new module format supports hot reload among other load-time optimizations.

If you encounter build failures, script errors, or loader issues after updating, you can opt back into the legacy AMD module format with the --module-format flag:
```
jaspr serve --module-format amd
```

See related change in `webdev`: https://github.com/dart-lang/webdev/pull/2839


Markzipan:
Note: this can land whenever, but ideally after this change (which is awaiting incoming hot restart changes): https://github.com/dart-lang/webdev/pull/2839

schultek (approved):
LGTM. 

Will merge this now. I hope to release hotreload soon as a beta version, but this would need https://github.com/dart-lang/build/pull/4945 to be resolved first.

Issue #816, opened 2026-06-04
jaspr serve causes 100% cpu usage


lewinpauli:
Hi, at first of all thank you a lot for creating such a great web framework for dart I am really not used to the web world and it helps a ton not to maintain npm and so many files while dart makes it so easy.

I just started with jaspr and I noticed that my Notebook is getting pretty hot (M1MacbookAir),

cpu utilization is at 100% (the whole time) for dart:server_target.dart

is that usual? 

I am using jaspr version: 0.23.1

Dart SDK version: 3.11.5 (stable)

lewinpauli:
I investigated this with the help of Claude and found the root cause.

**Bug chain:**

1. `jaspr_cli` generates `.dart_tool/jaspr/server_target.dart` with `debounceInterval: Duration.zero` ([dev_command.dart:464](https://github.com/schultek/jaspr/blob/main/packages/jaspr_cli/lib/src/commands/dev_command.dart))
2. `hotreloader` tries to watch `['bin', 'lib', 'test']` by default
3. For any path that doesn't exist, it creates a `PollingDirectoryWatcher(pollingDelay: _debounceInterval)` — and with `Duration.zero` as the delay, this becomes a tight busyloop ([hotreloader.dart:228-229](https://github.com/vegardit/dart-hotreloader/blob/main/lib/hotreloader.dart))

Most jaspr projects have neither a `bin/` nor a `test/` directory, so **two** tight-polling watchers spin indefinitely → 100% CPU even when no files change.

**Fix:** Simply creating both directories makes `hotreloader` use `DirectoryWatcher` (macOS FSEvents) instead of the polling loop. CPU drops to near-zero when idle.

```bash
mkdir -p bin test
```

A proper fix in `jaspr_cli` would be to use a non-zero `debounceInterval` (e.g. `Duration(milliseconds: 200)`) or to pass `excludedPaths: {'bin', 'test'}` to `HotReloader.create`.

PR #825, opened 2026-06-29
fix(cli): prevent hot reload server from pinning a CPU core (#816)

lewinpauli:
## Description

`jaspr serve` pins a CPU core at 100% even while completely idle (#816).

The generated `server_target.dart` created the `HotReloader` with
`debounceInterval: Duration.zero`. The `hotreloader` package reuses that value
as the `pollingDelay` for any watcher that falls back to polling — notably for
watch paths that don't exist, such as an unused `bin/` or `test/` directory.
With a zero delay this becomes a tight busy loop that pins a CPU core, on every
platform. Most Jaspr projects have neither `bin/` nor `test/`, so two such
watchers spin indefinitely.

This PR:

- Uses a non-zero `debounceInterval` (200ms) in the generated entrypoint, so any
  polling fallback no longer busy-loops. **This alone fixes the reported issue.**
- *(Optional optimization)* Additionally passes `excludedPaths` for `bin`/`test`
  when those folders are absent, so no polling watcher is created for them in the
  first place. Not strictly required once the above is in place — happy to drop
  it if you prefer a more minimal change.

## Type of Change

- 🛠️ Bug fix

## Ready Checklist


schultek:
Thanks for this fix.

Regarding excludePaths, the hotreloader docs say:
> [excludedPaths] are always relative to the project's root directory. If the project is a pub workspace, the `bin` folder of the package will be located in the `packages/<package>` folder. That means it should be specified like `packages/<package>/bin` in [excludedPaths], where `<package>` is the name of the package where the `HotReloader` is created. If the project consists of a single package, the `bin` directory lies in the project's root folder and should be specified like `bin` in [excludedPaths].

So we would have to have a more complex setup for these paths to detect if jaspr is run in a workspace or not.

To avoid that let's just drop the excluded paths, since you said the debounce duration already fixes it. I'm also ok with setting this higher, e.g. 1 second which is the default value.

lewinpauli:
Sorry for the delay, this slipped off my radar for a few weeks.

Fair enough on both counts. And good point about the workspace paths, I'd used bare
"bin"/"test" which wouldn't have matched in a workspace anyway.

Dropped "excludedPaths" and set "debounceInterval" to 1 second. The debounce is what
actually fixes it, since hotreloader reuses that value as the polling delay, so
"Duration.zero" was the real culprit. Just one line of change now.

schultek (approved):
LGTM. I just shortened the changelog entry a bit.

PR #833, opened 2026-07-20
[cli] Fix repeated copying of nested build assets

parlough:
Before this change, every directory was listed recursively, and all descendants were added back to the pending list. When those subdirectories were processed, their descendants were recursively added again, potentially causing repeated copies.

### DIFF
--- a/packages/jaspr_cli/lib/src/commands/base_command.dart
+++ b/packages/jaspr_cli/lib/src/commands/base_command.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:collection';
 import 'dart:convert';
 import 'dart:io';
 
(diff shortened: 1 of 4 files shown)


packages/jaspr_cli/lib/src/commands/base_command.dart
```diff
@@ -237,21 +237,32 @@ abstract class BaseCommand extends Command<int> {
...
     final to = project.requireMode != JasprMode.server ? 'build/jaspr' : 'build/jaspr/web';
+
+    // Pending relative paths to copy or inspect while traversing the source tree.
     final moveTargets = [...targets];
+    final visitedTargets = <String>{};
 
     final moves = <Future<void>>[];
     while (moveTargets.isNotEmpty) {
-      final moveTarget = moveTargets.removeAt(0);
+      final moveTarget = p.normalize(moveTargets.removeLast());
```
schultek: We should keep removing from the start, otherwise this isn't a queue.

schultek (approved):
LGTM

PR #834, opened 2026-07-22
feat: make sitemap priority optional (#785)

### LINKED ISSUE #785: feat: make `priority` optional for sitemap generation
**Description**

This [article from Google back in 2023 states that neither `changefreq` or `priority` is used](https://developers.google.com/search/blog/2023/06/sitemaps-lastmod-ping) by Google.

This [other document about sitemaps also stated the same that neither `changefreq` or `priority` is used](https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap#additional-notes-about-xml-sitemaps) and instead `lastmod` is.

It would be good if `priority` was optional like the way `changefreq` already is when building the sitemap.

bhavit04:
As discussed in #785, `priority` isn't used by search engines anymore, so it makes sense to be able to leave it out of the sitemap like `changefreq` already can be.

`RouteSettings.priority` now defaults to `null` instead of `0.5`, and the sitemap generator only writes the `<priority>` tag when it's set. Went with the default-null direction from the thread to keep it consistent with `changeFreq`.

Small heads-up: this makes `RouteSettings.priority` nullable, so it's technically breaking for anyone reading it as non-null. Updated the existing sitemap test to cover the omit case, and it passes.

Closes #785

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased patch
+
+- The sitemap `priority` tag is now omitted when a route has no priority set, instead of always defaulting to `0.5`.
+
 ## 0.23.2
 
 - Added `basePath` property to `AppBinding` to support hosting applications under a sub-path.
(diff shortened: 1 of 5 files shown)


schultek (approved):
LGTM

Fyi doing a minor version increase on a breaking change is actually fine on a 0.x version, since there minor version bumps are considered breaking anyways.

Issue #736, opened 2026-01-28
feat: Add `jaspr new component` command to generate components (and other things)


zeykafx:
**Description**

Creating new Jaspr components in VSCode is quite simple thanks to snippets. However, as mentioned in issue #651, some VSCode users are moving to Zed or other editors. Creating and maintaining a zed extension might not be on the roadmap right now, so this makes having a CLI command for component generation (and a lot of other potential things) a valuable addition. It would provide users of any editor a simple way to create components, routes, documentation pages, tests,… The idea came from https://dart-frog.dev/basics/routes/#creating-routes.

**Proposed command:**

My proposal is to add a new command that generates a stateless or stateful component (or other) based on a mason template. The command could look like this:

```bash
jaspr new component ContactInfo // stateless by default

jaspr new component TodoList --statefull
```

Using the `new` keyword would allow more templates to be generated by the command in the future (e.g., `jaspr new page` or `jaspr new route`, `jaspr new test`, …)

There is the possibility of new users getting confused with `jaspr create`, so perhaps another name might be better? Maybe `jaspr add` or `jaspr make`?

The choice between stateless and stateful components would be done using flags. This would allow the addition of other flags later on, such as `--async` for `AsyncStatelessComponent` or `--flutter` for `FlutterEmbedView`, or even other kinds of flags like `--with-test` or `--client`.

I think this could ease the learning curve and feature discovery for new jaspr users. At the very least, it would make using Jaspr with editors like Zed a bit better.

I’m willing to work on this if you believe that this would be a useful addition!

schultek:
I like the general idea, need to think a bit about it.

I was anyways thinking about adding a cli way to add AI rules and skills to the project, maybe this can be combined.

zeykafx:
Do you think the install-skills command could be folded into the "new" command? I have a working implementation of the "new" command that uses mason bricks for the "component" subcommand. A new "skills" subcommand could keep using the existing skills you created. What do you think?

PR #838, opened 2026-07-23
feat: jaspr_cli "new component" command

zeykafx:
## Description

This PR implements the `jaspr new` command as proposed in #736.

The command can generate stateless, stateful, async, inherited or Flutter embedded view components from a mason brick (There is one brick per component type), but also `jaspr_content` pages:

New component creation examples:

```shell
jaspr new component ContactInfo          # stateless by default
jaspr new component TodoList --stateful
jaspr new component LoadedData --async
jaspr new component Header --with-test   # also generates test/components/header_test.dart
jaspr new component AppTheme --inherited # creates an InheritedComponent under lib/

# Embeds a flutter widget, and sets the project up for flutter embedding
jaspr new component CounterView --flutter --flutter-app-name CounterWidget --with-flutter-widget

# Dry run: preview changes without touching the disk
jaspr new component CounterView --flutter --dry-run
```

Options:

- `--path` (`-p`): where the component is created, defaults to `./lib/components` (or `./lib` for inherited components)
	- Note that if the command is ran from somewhere outside of the project root then the component will be created in the current working directory.
	- For example, if the command is ran from "lib/components/blog", then the component is created there. If a path is provided, the component will be created under this path, e.g., running "jaspr new component ContactSheet --path contact" from "lib/components/blog" will create the component under "lib/component/blog/contact"
- `--stateless` / `--stateful` / `--async` / `--inherited` / `--flutter`: component type, one allowed, defaults to stateless
- `--client`: makes it a client component (ignored with a warning for async and inherited components)
- `--with-styles`: adds a `css` styles getter (ignored with a warning for inherited components)
- `--with-test`: generates a default test for the newly generated component, see below (not supported yet for inherited and flutter components, it is ignored with a warning)
- `--dry-run`: prints what would be generated without generating the files
- `--flutter-app-name` (or `--app-name`, and others): name of the Flutter widget to embed, only used with `--flutter`. If it is not given, the command prompts for it (defaults to `MyFlutterApp`).
- `--with-sample-flutter-widget` (or `--with-flutter-widget`, ...): also generates a sample Flutter widget with the name of the flutter app, only used with `--flutter`. The flag is negatable, and if it is not passed at all then the command prompts for it.

The component name can be given in PascalCase, camelCase or snake_case (the name will be converted by mason). The name is also checked to be a valid dart identifier after being converted.

NOTE: `--with-test` will create a new test file under the "test" directory, the file location mirrors the component's location in `lib/` but in `test/`. E.g.,  Creating a component`lib/components/blog/about_us.dart` will create a test at `test/components/blog/about_us_test.dart`. 
The test uses `testComponents` from `jaspr_test` (or `testServer` for `AsyncStatelessComponent`) and imports the component via using `package:xyz/...`. 
If `jaspr_test` is not found in the dev dependencies, then the command prompts the user to add it.

## Inherited components

`--inherited` generates an `InheritedComponent` instead of a regular component. The generated class has a sample data field with a TODO to indicate the user to replace the field, an `of` and a `maybeOf` static method, and an `updateShouldNotify`

Example:

```shell
jaspr new component AppTheme --inherited  # creates lib/app_theme.dart
```

NOTE: inherited components are created under `lib/` and not `lib/components/`, since they are not really component that will be rendered
Other note: `--client`, `--with-styles` and `--with-test` are ignored with a warning for these

## Flutter embedded view

Adding a FlutterEmbedView to a project requires multiple steps to be done (by hand today), but the command does the following:

- If no Flutter app name was passed with the `--flutter-app-name` option, then it prompts the user for one
- prompts to add `flutter` and `jaspr_flutter_embed` to the dependencies, and `flutter_lints` and `flutter_test` to the dev dependencies, only if they are missing
- sets `flutter: embedded` in the `jaspr` block of the pubspec file (if the flutter mode is already set to `plugins`, it prompts the user if they want to overwrite it or not)
- sets `uses-material-design: true` in the `flutter` block of the pubspec
- creates `web/flutter_bootstrap.js`
- includes a ref to that script where it belongs for the current rendering mode: in the head of the `Document` found in `lib/main.server.dart` for server and static, or in `web/index.html` for client. If the mode cannot be read, it warns the user and gives info on how to include it manually.
- generates a `FlutterEmbedView.deferred` component, with the `@Import.onWeb` ref to the widget and the `ViewConstraints`
- If the user passed `--with-sample-flutter-app`, it generates a sample Flutter widget under `lib/widgets/`, otherwise it prompts the user if they want a sample widget

Examples:

```shell
# prompts for the widget name and for the sample widget
jaspr new component CounterView --flutter

# no prompts
jaspr new component CounterView --flutter --app-name CounterWidget --with-flutter-widget
jaspr new component CounterView --flutter --app-name CounterWidget --no-with-flutter-widget
```

NOTE: the generated component is very similar to what `jaspr create` produces when flutter embedding is selected
NOTE: `--dry-run` is quite useful here since this command modifies quite a few things: it lists every change (deps, pubspec blocks, bootstrap script, script ref, generated files) without writing anything.

The command behaves in a similar manner to angular's `ng generate component [name] [options]` command.

## Content pages

`jaspr new page <name>` generates a new `jaspr_content` page from a mason brick, with the frontmatter filled in from the options.

Examples:

```shell
jaspr new page getting_started            # creates content/getting_started.md
jaspr new page guides/deploying           # nested, creates content/guides/deploying.md
jaspr new page guides --index             # creates content/guides/index.md
jaspr new page changelog --format html    # creates content/changelog.html
jaspr new page about --no-frontmatter     # no frontmatter block at all
```

Options: `--path` (`-p`), `--format` (defaults to md), `--layout`, `--title` (defaults to the name in Title Case), `--description`, and the flags `--dry-run`, `--index`, `--sitemap`, `--sitemap-exclude`, `--meta` and `--frontmatter` (can be negated, it is on by default).

The command also reads the `ContentApp` of the server entrypoint to check a few things:

- it reads the `directory` argument, so that the page is created in the right place when it is not `content` (i.e., if the user has changed the content directory and has not provided a path for the new page, the command will place it in the custom directory instead of `content/`)
- it checks `parsers` to prompt the user when the requested format might not be parsable (e.g., `--format html` while only `MarkdownParser()` is registered. Note that if the user wants another format than mdx/md/html this is ignored, of if the user has a custom parser we also ignore this warning)
- Like for the parsers, it checks `layouts` to prompt the user when the requested `--layout` may not be registered

NOTE: since jaspr_content can't run as a client side application, the command errors out for client projects. It also prompts to add `jaspr_content` to the dependencies when it is missing.

## VSCode extension support

The VSCode extension was updated to make use of the new CLI command to allow users to create new components or pages very easily.
The command is reachable from the command palette with "Jaspr: New component or page" and is split in two parts:

1. **New components**:
  - _Component type_: the user can pick from a list of pre-made component types (e.g., Stateless component, Stateful,...). A "more" option is present, and allows the user to pick from a list of more component types (e.g., "Stateful client components with tests",...)
  - _Component location_: The command provides 3 options: `lib/components`, `lib, or another directory (which the user browses to pick)
  - _Component name_: The name must be a valid dart identifier (the extension checks to avoid having the CLI throw the error)
2. **New pages**:
  - _Page type_: As for components, a list of premade page types can be selected (e.g., "Plain markdown page", "HTML blog", "Markdown docs",...). There is a custom option at the bottom which is described later
  - _Page file name_.
  - _Page location_: similar to components, provides 3 choices `contents`, `lib`, or another directory
  - For custom pages only:
    - _Page format_: lets the user pick one from MD/MDX/HTML/custom
    - _Page layout_: lets the user pick one from Blog/Docs/Empty/custom
    - _Page options_: Multi select picker where the user can select multiple options from `index`, `meta`, `frontmatter`, `sitemap`, `sitemap-exclude` (the extension warns the user (and exits) if they pick both `sitemap` and `sitemap-exclude` at the same time).
    - _Optional title and description_: can be skipped, the title defaults to the page name but in title case.

## Type of Change

## Ready Checklist


zeykafx:
@schultek This is still a WIP, but I'd love some feedback on this. Do you think it could be used to generate the skills too? Do you think there are more things that it should generate? Should `new` be changed to `generate` (to be more similar to angular's command)?

Thanks!!

schultek:
Looking forward to this.

> Do you think it could be used to generate the skills too?

I prefer to keep the current install-skills command and keep this separate for now.

zeykafx:
@schultek would it be better to try to integrate this directly into the `next` branch?

zeykafx:
The VSCode extension now supports the `new component` and `new page` commands. I based the code very largely on the `create` extension code.  I updated the description above to include more information.

The extension currently doesn't prompt the user if they want to add missing dependencies and instead raises a flag to tell the CLI to add them directly, I was debating just warning the user that they must include the missing deps themselves but that didn't seem like it would be the outcome expected by a user.

zeykafx:
@schultek This PR is quite large, would you prefer if I split it in three parts? 
1 for the new components command, then 1 for new pages, and 1 for the VSCode extension support. I could also target the next branch (because this doesn't use analyzer version 13 yet). 
Let me know what you prefer

schultek:
No need to split this up. But targeting the next branch would be great.

Also fyi I probably won't get around to properly reviewing this for a few weeks due to vacation.

zeykafx:
The CI formatting check fails due to duplicate "analyzer" keys in `packages/jaspr/analysis_options.yaml`, this is fixed in https://github.com/schultek/jaspr/pull/872. Other than that, this pr is now up to date with `next`

PR #844, opened 2026-08-01
Avoid empty classes in `Component.apply`

parlough:
Update class string splitting to handle extra middle or trailing whitespace (of different types).

Optionally we can still only split by normal spaces and still remove the empty entries. Let me know what you think!


schultek (approved):
LGTM

PR #845, opened 2026-08-01
Deduplicate client component parameter decoding

parlough:
The `params` field wasn't used as its logic was duplicated inline. It could have gotten out of date though as `data` can change, so instead change it to a getter and consolidate to using the getter.


schultek (approved):
LGTM

PR #853, opened 2026-08-04
fix(cli): normalize Windows CSS glob paths

LeonMollNetlight:
## Summary

Normalize the absolute build-directory path before using it as a CSS runner glob pattern.

## Root cause

On Windows, `Directory.path` uses backslashes. `package:glob` expects `/` in a glob pattern, so the drive-letter path was treated as relative and `jaspr serve` failed with `PathNotFoundException` after building web assets.

## Impact

`jaspr serve` with standalone CSS generation now works on Windows when using the CLI's CSS runner.

## Validation

- Ran `dart run jaspr_cli:jaspr serve --verbose` in DartPad's `dartpad_frontend` consumer on Windows.
- Confirmed the build daemon and CSS runner remained active; the previous `PathNotFoundException` did not occur.
- `dart pub get` in a fresh Jaspr checkout did not complete within the available time, so the package-local integration test was not run.

### DIFF
--- a/packages/jaspr_cli/lib/src/helpers/css_helper.dart
+++ b/packages/jaspr_cli/lib/src/helpers/css_helper.dart
@@ -337,7 +337,7 @@ class CssRunner {
       return <String>[];
     }
 
-    final runnerFiles = await Glob('${buildDir.path}/**.styles.dart').list().toList();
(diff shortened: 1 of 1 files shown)


schultek (approved):
LGTM

PR #855, opened 2026-08-04
Fix repeated lines during jaspr create rendering mode prompt

zeykafx:
## Description

This fixes a rendering bug when using `jaspr create` where the rendering mode prompt would be re-printed on every key press. This is due to a bug in `mason_logger` (see https://github.com/felangel/mason/issues/1356) that causes the list of choices to be reprinted to the terminal if the terminal has scrolled. A workaround is to clear the entire terminal, however this is not the behavior expected by users. The workaround here scrolls down enough to make space for the choices, then moves back up. 

Before (pressed down arrow twice): 
CleanShot 2026-08-04 at 12  36 28@2x
After (pressed down arrow twice):
CleanShot 2026-08-04 at 12  37 01@2x

## Type of Change

- 🛠️ Bug fix

## Ready Checklist

### DIFF
--- a/packages/jaspr/CHANGELOG.md
+++ b/packages/jaspr/CHANGELOG.md
@@ -7,6 +7,7 @@
 - The sitemap `priority` tag is now omitted when a route has no priority set, instead of always defaulting to `0.5`.
 - Fixed a misconfiguration of server hotreload that caused `jaspr serve` consuming excessive CPU.
 - Require `build_daemon ^4.1.4` to fix build error due to version mismatch.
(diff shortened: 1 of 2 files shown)


zeykafx:
Ah, I see that the `next` branch fixed this issue. @schultek when do you reckon `next` will be merged into main? I fixed another bug where the build daemon error logs wouldn't show up, but I suppose that making a PR for that is not useful then?

schultek:
Yes this is already fixed on next, which I can hopefully merge soon(ish). So I will close this one.

PR #856, opened 2026-08-05
Require Dart 3.13 or later

parlough:
- Update packages in the workspace to require Dart 3.13 or later.
  - Reformat with formatter changes.
  - Fix lints newly triggered as a result.
- Update pinned dependencies to account for the new constraints.

This will enable use of up-to-date build packages for implementing the hotreload changes.


schultek (approved):
LGTM

PR #857, opened 2026-08-05
Update melos and use workspace-version consistently

parlough:
- Update melos to v8.
  - Remove the no-longer needed dependency override of `cli_util`.
  - Remove the custom analyze script since melos now has `melos analyze`.
- Consistently use and suggest the workspace-vendored version of `melos` rather than a mix of it and a globally installed version.


.github/workflows/test.yml
```diff
@@ -38,15 +38,15 @@ jobs:
...
+          dart run melos bootstrap
           jaspr --disable-analytics
 
       - name: Check formatting
-        run: melos run format
+        run: dart run melos run format
 
       - name: Check linting
-        run: melos run analyze
+        run: dart run melos analyze
```
schultek:
```suggestion
        run: dart run melos analyze --fatal-infos
```

schultek (approved):
LGTM

PR #823, opened 2026-06-19
feat(analyzer)!: bump min analyzer version to ^13.0.0 and migrate AST usages

kevmoo:
Bump minimum required `analyzer` version lower bound to `^13.0.0` across workspace packages.

### Changes
- **AST Usages**: Migrated argument inspection in `jaspr_cli` migration visitors and `jaspr_lints` rules/assists from `NamedExpression` and `NodeList<Expression>` to analyzer v13 `NamedArgument` and `NodeList<Argument>`.
- **Warning Fix**: Fixed unawaited future return warning in `proxy_helper.dart`.
- **Test Matrix**: Excluded invalid `client` mode + `multi-page` routing combination from `jaspr_cli` integration test matrix.
- **Changelogs**: Documented `analyzer` `^13.0.0` support.

### Breaking Change
Minimum required `analyzer` version constraint is now `^13.0.0`.

PR #866, opened 2026-08-10
[next] Update to analyzer v13 and allow v14

parlough:
This bump is necessary to pull in the latest versions of `package:build_web_compilers` with relevant fixes. We can require v13 and allow v14 as v14 only removes previously deprecated members we don't rely on after this migration.

Supersedes and closes https://github.com/schultek/jaspr/pull/823

### DIFF
--- a/packages/jaspr_cli/lib/src/migrations/build_method_migration.dart
+++ b/packages/jaspr_cli/lib/src/migrations/build_method_migration.dart
@@ -1,6 +1,7 @@
 import 'dart:math';
 
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:io/ansi.dart';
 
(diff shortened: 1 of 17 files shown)


packages/jaspr_cli/lib/src/migrations/build_method_migration.dart
```diff
@@ -282,33 +282,37 @@ class BuilderVisitor extends RecursiveAstVisitor<void> {
...
-              name: 'Builder' ||
-                  'StatefulBuilder' ||
-                  'AsyncBuilder' ||
-                  'ListenableBuilder' ||
-                  'StreamBuilder' ||
-                  'FutureBuilder',
+    if (node.parent
+        case FunctionExpression(
+          parent: NamedArgument(
+            name: final name,
```
schultek: why not 

```suggestion
            name: Token(lexeme: 'builder')
```

schultek (commented):
This supersede #823 right?

parlough:
Thanks for taking a look @schultek :D

> This supersede https://github.com/schultek/jaspr/pull/823 right?

I didn't notice that PR, but yes. This once targets `next` and has some other changes, so let's go with this one. I updated the PR description to note this.

---

I made the simplification you suggested and completed a bunch of other similar clean up in the same vein, mostly around reducing duplication.

schultek (approved):
LGTM

PR #868, opened 2026-08-12
Fail the render instead of hanging when an initial build step throws

spydon:
## Description

If a step of the initial build throws from anywhere other than a component's `build` method, static rendering **hangs forever** instead of failing.

`AsyncBuildOwner.performRebuildOn` runs each step inside a `TaskChain`:

```dart
final chain = TaskChain.start()
    .then(() => child.performRebuild())
    .then(() => child._asyncBuildLock)
    .then(() => parentAsyncBuildLock)
    .then(() => child.didRebuild());
```

Nothing catches a throw from those callbacks. When one escapes, `c._complete()` is never reached, so that chain stays uncompleted forever. `completeInitialBuild` is awaiting `asFuture` on it, so it never calls `buildCallback`, the post-frame callback never fires, `ServerAppBinding.render` never returns, and no HTTP response is ever produced.

Since `jaspr build` has no timeout of its own, the consequence is not a failed build but a stuck one. In our case a release job sat on the build step for 30+ minutes before we cancelled it, and it would otherwise have run to the CI provider's 6-hour job limit:

```plaintext
✓ Completed building web assets. (62.4s)
Preparing static rendering...
✓ Server started (7.0s)
Generating routes...
(1/1) Generating route "/" ...
(2/14) Generating route "/schedule" ...      <- never returns
```

Errors thrown *inside* `build` are already handled — the `onError` handlers in `StatelessElement`/`StatefulElement.performRebuild` call `failRebuild`, which reports the error and yields a 500. I verified that path still behaves correctly. This PR closes the gap for everything else.

### Change

`TaskChain` records a failure, skips its remaining steps, and carries the failure to the end of the chain. `completeInitialBuild` then reports it through `AppBinding.reportBuildError` and completes the build regardless. The result is a reported error and a 500 response rather than a hang.

`asFuture` deliberately still completes normally in both cases. Completing it with an error would convert an uncaught build failure into an unhandled asynchronous error, which is precisely how this failure mode stayed invisible.

I also added a `if (_done) return;` guard to `_complete`, so a chain cannot be completed twice and have its listeners run twice.

### Tests

`packages/jaspr/test/server/build_error_test.dart` covers three cases, using an element that throws from `didRebuild` to fail a step outside the guarded `build` path:

| Case | Before | After |
|---|---|---|
| build step throws | **hangs** (test times out) | 500 |
| build step throws under a `PreloadStateMixin` ancestor | 500 | 500 |
| nothing throws | 200 | 200 |

Confirmed the first test times out with `render never completed` when the fix is reverted, and passes with it. Full suite: **212 passing** (209 pre-existing + 3 new), on VM and Chrome.

### Relationship to #867

#867 touches the same file and is independent — it swaps `TaskChain._listeners` from a `Set` to a `List` to work around a Dart VM soundness fault (a non-nullable `int` observed as null inside `_LinkedHashSetMixin.add`). They will conflict textually; happy to rebase whichever lands second.

The two are worth having separately: that VM fault is what threw inside the chain machinery for us, and *this* bug is why it cost us a stuck 6-hour job instead of a red build.

## Type of Change

- 🛠️ Bug fix

## Ready Checklist

### DIFF
--- a/packages/jaspr/test/server/build_error_test.dart
+++ b/packages/jaspr/test/server/build_error_test.dart
@@ -1,0 +1,84 @@
+@TestOn('vm')
+library;
+
+import 'package:jaspr/dom.dart';
+import 'package:jaspr/server.dart';
+import 'package:jaspr_test/server_test.dart';
+
+import 'render_test.dart';
+
+void main() {
+  group('build error', () {
+    setUpAll(() {
+      Jaspr.initializeApp();
+    });
+
+    test('completes the render when a build step throws outside build()', () async {
+      final result = await renderComponent(const ThrowsInDidRebuild()).timeout(
+        const Duration(seconds: 30),
+        onTimeout: () => fail('render never completed'),
+      );
+
+      expect(result.statusCode, 500);
+    });
+
+    test('completes the render when the throwing element has an async ancestor', () async {
+      final result = await renderComponent(const Preloading(child: ThrowsInDidRebuild())).timeout(
+        const Duration(seconds: 30),
+        onTimeout: () => fail('render never completed'),
+      );
+
+      expect(result.statusCode, 500);
+    });
+
+    test('renders normally when nothing throws', () async {
+      final result = await renderComponent(const Preloading(child: Component.text('ok')), standalone: true);
+
+      expect(result.statusCode, 200);
+      expect(result.body, decodedMatches(contains('ok')));
+    });
+  });
+}
+
+/// Throws from [Element.didRebuild], which [AsyncBuildOwner] invokes from inside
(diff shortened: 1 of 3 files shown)


schultek (approved):
LGTM

packages/jaspr/test/server/build_error_test.dart
```diff
@@ -0,0 +1,84 @@
+@TestOn('vm')
+library;
```
schultek, line 3:
```suggestion
/// @docImport 'package:jaspr/src/server/async_build_owner.dart';
@TestOn('vm')
library;
```

schultek (dismissed):

schultek (approved):

PR #873, opened 2026-08-21
[next] Remove unnecessary version checks

parlough:
Remove unnecessary version checks now that Dart 3.13 or later is required.


schultek (approved):
LGTM

PR #874, opened 2026-08-21
Require and use stable version of Dart 3.13

parlough:


schultek (dismissed):
LGTM. Can you add this also to the changelogs of the packages.

parlough:
> LGTM. Can you add this also to the changelogs of the packages.

@schultek Thanks for taking a look! I updated the changelogs.

However doing so dismissed your approval. Please take another look :)

schultek (approved):

PR #879, opened 2026-09-01
[cli] Use cli_util to locate Dart executable

parlough:
Takes advantage of the new top-level `dartExecutable` getter in the 0.6.0 release of `package:cli_util`: https://pub.dev/packages/cli_util/changelog#060

### DIFF
--- a/packages/jaspr_cli/pubspec.yaml
+++ b/packages/jaspr_cli/pubspec.yaml
@@ -58,6 +58,9 @@ dev_dependencies:
   test: ^1.22.0
 
 dependency_overrides:
+  # Some dependencies, such as `dart_data_home`, still require `cli_util: ^0.5.0`,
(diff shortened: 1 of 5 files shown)


packages/jaspr_cli/pubspec.yaml
```diff
@@ -58,6 +58,7 @@ dev_dependencies:
   test: ^1.22.0
 
 dependency_overrides:
+  cli_util: ^0.6.0
```
schultek, line 64: Why is this needed? Can we add a comment for this to know when we can remove it again.

schultek (dismissed):
LGTM

packages/jaspr_cli/pubspec.yaml
parlough: Added an explanatory comment. Some dependencies still use `0.5.0` but not in an incompatible way.

schultek (approved):

PR #880, opened 2026-09-08
[jaspr] Add child-list mutation and server-rendered hydration tests

parlough:
These will serve as a useful reference while I work on some fixes and optimizations to relevant logic.

- Adds regression tests for hydration/reload behavior using real server-rendered HTML.
  - Identifies known nested reload failures as skipped regressions with TODO comments.
- Adds regression tests for child-list mutations.
  - Adds TODOs for behavior that isn't well specified or guaranteed.
- Handles non-file script URIs needed by hybrid tests.


schultek (approved):
LGTM

PR #881, opened 2026-09-08
[next] refactor Component.apply handling

### LINKED ISSUE #859: [next] `Component.apply` properties lost on hydration
On branch `next`, when a client component applies properties to another client component with `Component.apply` when server rendered, it works but then the properties are lost on hydration. https://github.com/schultek/jaspr/tree/bug/client-component-apply/examples/client_component_apply_issue contains a runnable example, roughly like:

```dart
@client
class Outer extends StatelessComponent {
  const Outer({required this.child, super.key});

  final Component child;

  @override
  Component build(BuildContext context) => .apply(
    classes: 'highlighted',
    attributes: const {'data-outer': 'true'},
    child: child,
  );
}

@client
class InnerButton extends StatelessComponent {
  const InnerButton({super.key});

  @override
  Component build(BuildContext context) => button([
    .text('Click me'),
  ]);
}

void main() {
  Jaspr.initializeApp(
    options: defaultServerOptions,
  );

  runApp(
    const Outer(
      child: InnerButton(),
    ),
  );
}
```

In this example, the `<button>` has the applied `class="highlighted" data-outer="true"` at first but they're lost once hydrated.

schultek:
## Description

Refactors parts of how `Component.apply` is handled and works in combination with `SlottedChildView`.

The resolution of `.apply()` parameters is shifted to the render-object phase instead of the build phase by using a resolver callback in `_InheritedDomComponent` instead of storing the parameters directly. 

Fixes #859 

---

## Type of Change

- 🛠️ Bug fix

## Ready Checklist

### DIFF
--- a/packages/jaspr/lib/src/framework/components.dart
+++ b/packages/jaspr/lib/src/framework/components.dart
@@ -91,7 +91,7 @@ class DomElement extends DomRenderObjectElement {
     var attributes = component.attributes;
     var events = component.events;
 
-    if (inheritedDomParams case final inheritedDomParams?) {
+    if (inheritedDomParamsFor(renderObject) case final inheritedDomParams? when inheritedDomParams.isNotEmpty) {
       final classesSet = {...?_splitClasses(classes)};
       styles = Map.of(styles ?? {});
       attributes = Map.of(attributes ?? {});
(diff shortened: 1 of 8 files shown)


packages/jaspr/lib/src/framework/components.dart
```diff
@@ -227,57 +227,46 @@ abstract class DomRenderObjectElement extends MultiChildRenderObjectElement {
...
-        for (final child in buildOwnChildren()) _InheritedDomComponent(params: newParams, child: child),
-      ];
-    }
-    return buildOwnChildren();
+  @protected
+  Component wrapWithInheritedDomComponent({
+    required DomParamsResolver getParams,
+    required Component child,
+  }) {
+    return _InheritedDomComponent(getParams: getParams, child: child);
```
parlough: Should `_InheritedDomComponent` have a key that is set and derived if its child has one? That way if the `_InheritedDomComponent` wrapper of a child changes, the state is preserved?


```dart
    return _InheritedDomComponent(
      key: child.key == null ? null : ValueKey(child.key),
      getParams: getParams,
      child: child,
    );
```

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -300,27 +289,55 @@ class ApplyParams {
...
+
+      if (params.target.onlyChildren && !_isDirectDomChild(target, applyParent)) {
+        return parentParams;
+      }
+
+      if (parentParams.isEmpty) {
+        return [params];
+      }
+
+      return [...parentParams, params];
```
parlough, line 336: This seems the same unless you noticed some performance issue with it:

```suggestion
      return [...parentParams, params];
```

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -300,27 +289,55 @@ class ApplyParams {
...
+  }
+  return false;
+}
+
 class _InheritedDomComponent extends InheritedComponent {
-  const _InheritedDomComponent({required this.params, required super.child});
+  const _InheritedDomComponent({required this.getParams, required super.child});
 
-  final List<ApplyParams> params;
+  final DomParamsResolver getParams;
```
parlough: Maybe `resolveParameters` instead of `getParams`?

packages/jaspr/lib/src/framework/components.dart
```diff
@@ -300,27 +289,55 @@ class ApplyParams {
...
-      return _InheritedDomComponent(params: [params], child: child);
+    final parentResolver = context.dependOnInheritedComponentOfExactType<_InheritedDomComponent>()?.getParams;
+
+    List<ApplyParams> getParams(RenderObject target) {
+      final parentParams = parentResolver?.call(target) ?? const [];
+      final applyParent = (context as Element)._parentRenderObjectElement?.renderObject;
+
+      if (params.target.onlyChildren && !_isDirectDomChild(target, applyParent)) {
+        return parentParams;
+      }
```
parlough, line 330: I don't fully understand, but:

Now that the applied parameter resolution depends on render-object ancestry here, do the calls to `updateRenderObject` need to happen later? It seems, for example, when it’s called in `DomElement.didChangeDependencies`, the render object wouldn't be attached to its new parent yet.

parlough (commented):
Thanks for tackling this! I'll take a closer look and test it out soon, but a few initial comments:

packages/jaspr/lib/src/framework/components.dart
schultek: Yeah this is just a tiny optimization, can't hurt.

packages/jaspr/lib/src/framework/components.dart
schultek: Good catch, done

packages/jaspr/lib/src/framework/components.dart
schultek: Done

packages/jaspr/lib/src/framework/components.dart
schultek: Yes true, I refactored so that didChangeDependencies only sets a flag, and after rebuild, updateRenderObject happens after attachRenderObject so that the structure is clean when the parameters are resolved.

parlough (approved):
Thanks for those adjustments!

We might hit more cases but this seems the right direction.

PR #883, opened 2026-09-08
[next] improve dev server connection handling

### LINKED ISSUE #878: fix: dev server leaks one proxy connection per non-asset request (Cascade drops un-drained shelf_proxy responses)
**Description**

In dev (`jaspr serve`), every request that is *not* served by the webdev proxy leaks one TCP connection — and one un-drained HTTP response — to the proxy port, for the lifetime of the server. On a multi-page site that is one leaked connection per page view.

The cause is in `createHandler` ([`server_handler.dart`](https://github.com/schultek/jaspr/blob/main/packages/jaspr/lib/src/server/server_handler.dart)):

```dart
if (!kGenerateMode) {
  cascade = cascade.add(gzipMiddleware(staticHandler));   // proxyHandler -> JASPR_PROXY_PORT
}

cascade = cascade.add((request) async {  /* render the page */ });
```

A page path (`/about`) is not an asset, so the proxy answers 404 and `Cascade` moves on to the render handler. But `Cascade` **discards** the response it just cascaded past, and `shelf_proxy` builds that response around the raw `IOStreamedResponse.stream`:

```dart
return Response(clientResponse.statusCode,
    body: clientResponse.stream, headers: clientResponse.headers);
```

A discarded response is a body stream that is never listened to, so `HttpClient` never releases the socket it is riding on. Nothing in the chain drains it.

Interestingly `jaspr_cli` already avoids `Cascade` in its own proxy for a neighbouring reason — [`proxy_helper.dart`](https://github.com/schultek/jaspr/blob/main/packages/jaspr_cli/lib/src/helpers/proxy_helper.dart) carries the comment *"This is also the reason why `Cascade()` won't work here."*

**Steps To Reproduce**

Standalone, no jaspr involved — this is the `createHandler` shape reduced to `shelf` + `shelf_proxy`:

```dart
// pubspec: shelf ^1.4.0, shelf_proxy ^1.0.4
import 'dart:io';
import 'package:shelf/shelf.dart';
import 'package:shelf/shelf_io.dart' as io;
import 'package:shelf_proxy/shelf_proxy.dart';

Future<void> main() async {
  // Stands in for the webdev asset server: 404s for page paths.
  final upstream = await io.serve((_) => Response.notFound('nope'), 'localhost', 0);

  final app = await io.serve(
    Cascade()
        .add(proxyHandler('http://localhost:${upstream.port}/'))
        .add((_) => Response.ok('rendered page'))
        .handler,
    'localhost',
    0,
  );

  int upstreamConns() => (Process.runSync('lsof', ['-nP', '-i', ':${upstream.port}'])
          .stdout as String)
      .split('\n')
      .where((l) => l.contains('ESTABLISHED'))
      .length ~/ 2; // a local socket shows once per end

  final client = HttpClient();
  var done = 0;
  for (final target in [10, 50, 100, 200]) {
    while (done < target) {
      final req = await client.getUrl(Uri.parse('http://localhost:${app.port}/some/page'));
      await (await req.close()).drain<void>();
      done++;
    }
    await Future<void>.delayed(const Duration(milliseconds: 300));
    print('$done requests -> ${upstreamConns()} connections held to upstream');
  }
  exit(0);
}
```

```
10 requests -> 10 connections held to upstream
50 requests -> 50 connections held to upstream
100 requests -> 100 connections held to upstream
200 requests -> 200 connections held to upstream
```

Exactly 1:1, and they are never released.

On a real `jaspr serve` (static mode, `--port 8081`, proxy on the default 5567) the same thing shows up directly:

```
$ lsof -nP -i:5567 | grep -c ESTABLISHED        # 2 (idle)
$ for i in $(seq 1 400); do curl -s -o /dev/null http://localhost:8081/about; done
$ lsof -nP -i:5567 | grep -c ESTABLISHED        # 802  (lsof counts both ends)
```

Static asset requests (`/styles.css`) do **not** leak — those get a 200, the body is consumed, and the connection is reused. It is only the cascaded-past 404s.

Client-side symptom while this piles up: a steady stream of

```
ERROR - Asynchronous error
SocketException: Connection reset by peer (OS Error: Connection reset by peer, errno = 54), address = 0.0.0.0, port = 5567
```

**Doctor Output**

```
[✓] Jaspr CLI (Version 0.23.2)
  • Dart Version 3.13.2 (stable) on "macos_arm64"
  • Running on macos Version 26.6.2 (Build 25G83) - Locale en-US
[✓] Current Project
  • Dependencies on core packages:
    • jaspr: ^0.23.1
    • jaspr_builder: ^0.23.1 (dev)
    • jaspr_router: ^0.8.2
  • Rendering mode: static
  • Flutter mode: none
```

`createHandler` is byte-identical in jaspr 0.23.1, 0.23.4 and `main`, so this is not version-specific.

**Expected Behavior**

Falling through the proxy should release the proxied connection. Doing the fall-through by hand instead of via `Cascade`, and draining what gets dropped, fixes it:

```dart
final proxy = gzipMiddleware(staticHandler);

// ...instead of cascade.add(proxy):
final res = await proxy(request);
if (res.statusCode != 404) return res;
await res.read().drain<void>();   // release the proxied connection
// fall through to the render handler
```

Same repro with that change:

```
10 requests -> 1 connection held to upstream
50 requests -> 1 connection held to upstream
100 requests -> 1 connection held to upstream
200 requests -> 1 connection held to upstream
```

(One, reused via keep-alive.)

**Additional Context**

I did not measure a latency regression from this on its own — page renders stayed at ~4 ms with 2000 connections outstanding, and the sockets do get reclaimed eventually. The visible cost is the `Connection reset by peer` noise in the console and an ever-growing fd count on both the server and the `jaspr serve` process.

Found while chasing a separate 100% CPU problem, which turned out to be #816 (already fixed in jaspr_cli 0.23.3+).

schultek:
## Description

Fixes #878

## Type of Change

- 🛠️ Bug fix

## Ready Checklist

### DIFF
--- a/packages/jaspr/lib/src/server/server_handler.dart
+++ b/packages/jaspr/lib/src/server/server_handler.dart
@@ -85,7 +85,7 @@ Handler createHandler(
 
   // We skip static file handling in generate mode to always generate fresh content on the server.
   if (!kGenerateMode) {
-    cascade = cascade.add(gzipMiddleware(staticHandler));
+    cascade = cascade.add(gzipMiddleware(_drainingHandler(staticHandler)));
   }
 
   cascade = cascade.add((request) async {
@@ -124,7 +124,11 @@ Future<String?> Function(String) proxyFileLoader(Request req, Handler proxyHandl
       protocolVersion: req.protocolVersion,
     );
     final response = await proxyHandler(indexRequest);
-    return response.statusCode == 200 ? response.readAsString() : null;
+    if (response.statusCode == 200) {
+      return response.readAsString();
+    }
+    await response.read().drain<void>().catchError((_) {});
+    return null;
   };
 }
 
@@ -132,6 +136,19 @@ Handler createProxyHandler(http.Client? client) {
   final handler = proxyHandler('http://localhost:$jasprProxyPort/', client: client);
   // Determine and pass the base path to the proxy handler so it can rewrite DWDS handler paths correctly.
   return (req) => handler(req.change(headers: {'jaspr_base_path': req.handlerPath}));
+}
+
+/// Drains the response body for cascading status codes (404/405) before
+/// [Cascade] drops the response, preventing connection leaks when proxying.
+Handler _drainingHandler(Handler handler) {
+  return (request) async {
+    final response = await handler(request);
+    if (response.statusCode == 404 || response.statusCode == 405) {
+      await response.read().drain<void>().catchError((_) {});
+      return response.change(body: '');
+    }
+    return response;
+  };
 }
 
 // coverage:ignore-start
--- a/packages/jaspr/lib/src/server/server_handler.dart
+++ b/packages/jaspr/lib/src/server/server_handler.dart
@@ -143,7 +143,7 @@ Handler createProxyHandler(http.Client? client) {
 Handler _drainingHandler(Handler handler) {
   return (request) async {
     final response = await handler(request);
-    if (response.statusCode == 404 || response.statusCode == 405) {
+    if (response.statusCode case 404 || 405) {
       await response.read().drain<void>().catchError((_) {});
       return response.change(body: '');
     }
--- a/packages/jaspr/test/server/server_handler_test.dart
+++ b/packages/jaspr/test/server/server_handler_test.dart
@@ -1,0 +1,115 @@
+@TestOn('vm')
+library;
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:http/http.dart' as http;
+import 'package:jaspr/server.dart';
+import 'package:jaspr/src/server/server_handler.dart';
+import 'package:shelf/shelf_io.dart' as io;
+import 'package:shelf_proxy/shelf_proxy.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('server_handler', () {
+    setUpAll(() {
+      Jaspr.initializeApp();
+    });
+
+    test('drains proxy response when falling through to render handler', () async {
+      var streamListened = false;
+      var streamDrained = false;
+
+      // Create a mock handler that simulates a proxy returning 404 with a stream body.
+      final controller = StreamController<List<int>>(
+        onListen: () {
+          streamListened = true;
+        },
+      );
+      controller.add([1, 2, 3]);
+      unawaited(
+        controller.close().then((_) {
+          streamDrained = true;
+        }),
+      );
+
+      Response mockProxyHandler(Request request) {
+        return Response.notFound(controller.stream);
+      }
+
+      final handler = createHandler(
+        (request, render) {
+          return Response.ok('rendered page');
+        },
+        fileHandler: mockProxyHandler,
+      );
+
+      final response = await handler(Request('GET', Uri.parse('http://localhost/about')));
+      expect(response.statusCode, equals(200));
+      expect(await response.readAsString(), equals('rendered page'));
+
+      // Wait a tick to allow microtasks to finish.
+      await Future<void>.delayed(Duration(milliseconds: 50));
+
+      expect(streamListened, isTrue, reason: 'The 404 response body stream must be listened to');
+      expect(streamDrained, isTrue, reason: 'The 404 response body stream must be drained');
+    });
+
+    test('drains real shelf_proxy response on 404 fallthrough', () async {
+      // Upstream server stands in for dev asset server, returning 404 for non-asset requests.
+      final upstream = await io.serve((_) => Response.notFound('Not a static file'), InternetAddress.loopbackIPv4, 0);
+
+      final client = http.Client();
+      final proxy = proxyHandler('http://localhost:${upstream.port}/', client: client);
+
+      final handler = createHandler(
+        (request, render) {
+          return Response.ok('rendered page');
+        },
+        fileHandler: proxy,
+        client: client,
+      );
+
+      // Make multiple requests through the handler
+      for (var i = 0; i < 5; i++) {
+        final response = await handler(Request('GET', Uri.parse('http://localhost/about')));
+        expect(response.statusCode, equals(200));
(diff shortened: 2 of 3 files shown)


packages/jaspr/lib/src/server/server_handler.dart
```diff
@@ -149,6 +138,19 @@ Handler createProxyHandler(http.Client? client) {
...
 }
 
+/// Drains the response body for cascading status codes (404/405) before
+/// [Cascade] drops the response, preventing connection leaks when proxying.
+Handler _drainingHandler(Handler handler) {
+  return (request) async {
+    final response = await handler(request);
+    if (response.statusCode == 404 || response.statusCode == 405) {
+      await response.read().drain<void>().catchError((_) {});
+      return response.change(body: '');
```
parlough, line 148: Consider just creating a new `Response` here with the appropriate status code to avoid unnecessary work:

```suggestion
      return Response(response.statusCode);
```

packages/jaspr/lib/src/server/server_handler.dart
```diff
@@ -149,6 +138,19 @@ Handler createProxyHandler(http.Client? client) {
   return (req) => handler(req.change(headers: {'jaspr_base_path': req.handlerPath}));
 }
 
+/// Drains the response body for cascading status codes (404/405) before
+/// [Cascade] drops the response, preventing connection leaks when proxying.
+Handler _drainingHandler(Handler handler) {
+  return (request) async {
+    final response = await handler(request);
+    if (response.statusCode == 404 || response.statusCode == 405) {
```
parlough: If you're interested:

```suggestion
    if (response.statusCode case 404 || 405) {
```

packages/jaspr_cli/lib/src/helpers/proxy_helper.dart
```diff
@@ -29,16 +29,10 @@ mixin ProxyHelper on BaseCommand {
     final allowedFlutterPaths = RegExp(r'^assets|^canvaskit|^packages|.js$|.wasm$');
     final webdevHandler = devProxy?.handler ?? (req) => Response.notFound(null);
 
-    final cascade = Cascade().add(generatedHandler).add((req) async {
-      if (req.url.path == r'$jasprMessageHandler') {
-        onMessage?.call(jsonDecode(await req.readAsString()));
-        return Response.ok(null);
-      }
-
+    Future<Response> handleRequest(Request req, [Stream<List<int>>? body, bool isRetry = false]) async {
```
parlough: Consider making these named arguments. Reads much easier at the call site.

packages/jaspr/test/server/server_handler_test.dart
```diff
@@ -0,0 +1,115 @@
...
+      }
+
+      final handler = createHandler(
+        (request, render) {
+          return Response.ok('rendered page');
+        },
+        fileHandler: mockProxyHandler,
+      );
+
+      final response = await handler(Request('GET', Uri.parse('http://localhost/about')));
```
parlough:
```suggestion
      final response = await handler(Request('GET', Uri.http('localhost', '/about')));
```

packages/jaspr/test/server/server_handler_test.dart
```diff
@@ -0,0 +1,115 @@
...
+        (request, render) {
+          return Response.ok('rendered page');
+        },
+        fileHandler: proxy,
+        client: client,
+      );
+
+      // Make multiple requests through the handler
+      for (var i = 0; i < 5; i++) {
+        final response = await handler(Request('GET', Uri.parse('http://localhost/about')));
```
parlough:
```suggestion
        final response = await handler(Request('GET', Uri.http('localhost', '/about')));
```

packages/jaspr/test/server/server_handler_test.dart
```diff
@@ -0,0 +1,115 @@
...
+        controller.close().then((_) {
+          streamDrained = true;
+        }),
+      );
+
+      Response mockProxyHandler(Request request) {
+        return Response.notFound(controller.stream);
+      }
+
+      final loader = proxyFileLoader(Request('GET', Uri.parse('http://localhost/about')), mockProxyHandler);
```
parlough:
```suggestion
      final loader = proxyFileLoader(Request('GET', Uri.http('localhost', '/about')), mockProxyHandler);
```

packages/jaspr/test/server/server_handler_test.dart
```diff
@@ -0,0 +1,115 @@
...
+      final result = await loader('index.html');
+
+      expect(result, isNull);
+
+      await Future<void>.delayed(Duration(milliseconds: 50));
```
parlough: Since we require Dart 3.13 now:

```suggestion
      await Future.pause(Duration(milliseconds: 50));
```

packages/jaspr/test/server/server_handler_test.dart
```diff
@@ -0,0 +1,115 @@
...
+        },
+        fileHandler: mockProxyHandler,
+      );
+
+      final response = await handler(Request('GET', Uri.parse('http://localhost/about')));
+      expect(response.statusCode, equals(200));
+      expect(await response.readAsString(), equals('rendered page'));
+
+      // Wait a tick to allow microtasks to finish.
+      await Future<void>.delayed(Duration(milliseconds: 50));
```
parlough: Since we require Dart 3.13 now:

```suggestion
      await Future.pause(Duration(milliseconds: 50));
```

parlough (dismissed):
Nice. Looks good to me!

packages/jaspr/lib/src/server/server_handler.dart
schultek: Not sure about this, we will loose headers then.

parlough (approved):
Thanks for that additional clean up. Still looks good to me!

PR #887, opened 2026-09-12
Tighten framework class hierarchies

parlough:
To avoid exposing implementation or behavior that isn't useful or necessary, supporting future evolution of the framework.

If the hierarchies that were closed here need to be opened back up in the future, we can do so explicitly, documenting the constraints and expectations around them.

This is a step towards my work implementing some performance improvements to child node handling.


schultek (approved):
LGTM

Also thanks for structuring the changelog.

PR #888, opened 2026-09-12
Add render adapter and child-list range nesting tests

parlough:
Add additional tests, mostly around server handling of markup. Some are skipped with TODOs as they are failing due to framework issues.

These additional tests will help me validate some performance improvements I'm exploring.


schultek (approved):
LGTM

Do we need to make issues for the TODOs?

PR #889, opened 2026-09-12
Update `ChildNode.insertX` parameters to be non-nullable

parlough:
`null` is never passed nor does it make much sense to insert it.


schultek (approved):
LGTM

PR #890, opened 2026-09-12
Clean up `ChildList.wrapElement`

parlough:
Extract the boundary determination into a shared `encloses` local function to flatten the loops to make the logic easier to follow. Also add a basic API doc comment.

### DIFF
--- a/packages/jaspr/lib/src/server/child_nodes.dart
+++ b/packages/jaspr/lib/src/server/child_nodes.dart
@@ -182,60 +182,40 @@ class ChildList with Iterable<MarkupRenderObject> {
     return null;
   }
 
+  /// Wraps [element]'s rendered output in a new range and returns it.
+  ///
+  /// The output must already be a direct child of this list,
+  /// not nested in a fragment.
+  ///
+  /// Ranges around the same output nest by element depth,
+  /// so ancestor ranges enclose descendant ranges.
+  /// For the same element, higher [priority] ranges enclose lower ones,
+  /// and at equal priority the newest range is outermost.
   ChildListRange wrapElement(Element element, [int priority = 0]) {
     final node = find(element.slot.target!.renderObject as MarkupRenderObject);
     assert(node != null, 'Element not found in child list');
 
-    ChildNode startBefore = node!;
-    ChildNode endAfter = node;
-
-    while (true) {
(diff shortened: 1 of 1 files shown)


packages/jaspr/lib/src/server/child_nodes.dart
```diff
@@ -182,60 +182,40 @@ class ChildList with Iterable<MarkupRenderObject> {
...
     final node = find(element.slot.target!.renderObject as MarkupRenderObject);
     assert(node != null, 'Element not found in child list');
 
-    ChildNode startBefore = node!;
-    ChildNode endAfter = node;
+    /// Whether the new range should enclose the adjacent existing [boundary].
+    bool encloses(ChildNodeBoundary boundary) {
+      if (boundary.element == element) return priority >= boundary.priority;
+      assert(boundary.element.depth != element.depth);
+      return element.depth <= boundary.element.depth;
```
schultek, line 202: Just adding some comments to make this more understandable.

```suggestion
      // When wrapping the same element, higher priority wraps lower priority.
      if (boundary.element == element) return priority >= boundary.priority;
      assert(boundary.element.depth != element.depth);
      // Outer elements wrap inner elements.
      return element.depth <= boundary.element.depth;
```

packages/jaspr/lib/src/server/child_nodes.dart
```diff
@@ -182,60 +182,40 @@ class ChildList with Iterable<MarkupRenderObject> {
...
-        } else {
-          // Element is an ancestor of startPrev, incorrect order
-          startBefore = startPrev;
-          continue;
-        }
-      }
-      break;
+      final prev = startBefore.prev;
+      if (prev is! ChildNodeBoundary || prev.range.start != prev || !encloses(prev)) break;
+      startBefore = prev;
```
schultek, line 210: Super nit, but I find it more intuitive if we turn this around
```suggestion
      if (prev is ChildNodeBoundary && prev.range.start == prev && encloses(prev)) {
        startBefore = prev;
        continue;
      }
      break;
```

schultek (dismissed):
LGTM, just some nit suggestions

schultek (approved):
LGTM, thanks for adjusting

